PR #21, opened 2015-04-23
Docker

dshimola:
What you will need to do to test this is install Docker (https://docs.docker.com/installation) and docker-compose (https://docs.docker.com/compose/install/). 
Once you have installed docker and compose then navigate to the /data/server/docker directory. 
Ensure that the ports that docker containers will use are not already taken (port 80 and 4022).
After ensuring that the ports are free, run ' docker-compose up -d '
Note the password that is displayed after step 2 or so (The output will look like "password is 14characterpassword" This password is only show on the initial creation of the images. 

king-phisher is now running, you can connect to it through the client with (from localhost, use the docker interface ip) and ssh port 4022 using the username 'kingphisher' and password that was given during the initial docker-compose up. 

to stop the images, use docker-compose stop

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ DODO
 todo.txt
 
 data/client/king_phisher/#king-phisher-client.ui#
+!data/server/docker/docker-compose.yml
 !data/server/king_phisher/error_*.html
 !data/server/king_phisher/server_config.yml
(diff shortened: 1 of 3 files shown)


zeroSteiner:
This all looks good and I've tested it with an Ubuntu 14.04 system. Will merge this in momentarily.

PR #30, opened 2015-07-06
Added Metasploit plugin that uses King Phisher's REST API

coldfusion39:


zeroSteiner:
This is awesome, can't wait to test it out.

data/msf/sms.rb
```diff
@@ -0,0 +1,186 @@
...
+      self.framework.events.remove_session_subscriber(@inst)
+      remove_console_dispatcher('sms')
+    end
+
+    def name
+      "sms"
+    end
+
+    def desc
+      "Sends a SMS message when recieving a new session through the use of the King Phisher's REST API."
```
zeroSteiner: Static strings like this should use single quotes because they are not using interpolation, ie: ```'this is a static string'``` vs ```"this string uses interpolation to print #{some_variable}"```.

data/msf/sms.rb
```diff
@@ -0,0 +1,186 @@
...
+# SOFTWARE.
+
+require 'uri'
+module Msf
+
+  class Plugin::SessionSMS < Msf::Plugin
+    include Msf::SessionEvent
+
+    if not defined?(Sms_yaml)
+      Sms_yaml = "#{Msf::Config.get_config_root}/sms.yaml"
```
zeroSteiner: This would probably be better off as a method like to maintain the snake_case convention:
```
def sms_yaml
  "#{Msf::Config.get_config_root}/sms.yaml"
end
```

data/msf/sms.rb
```diff
@@ -0,0 +1,186 @@
...
+        end
+      end
+
+      def cmd_sms_stop
+        print_good("Stopping SMS alerting!")
+        self.framework.events.remove_session_subscriber(self)
+      end
+
+      def cmd_sms_save
+        if @king_phisher_server and @king_phisher_token and @sms_number and @sms_carrier
```
zeroSteiner: In ruby it's almost always better to use the binary operators ```&&``` and ```||``` instead of the word operators like "and" and "or".

data/msf/sms.rb
```diff
@@ -0,0 +1,186 @@
...
+        if args.length > 0
+          print_status("Setting King Phisher's REST API token to #{args[0]}")
+          @king_phisher_token = args[0]
+        else
+          print_error("Please provide the REST API token of your King Phisher server!")
+        end
+      end
+
+      def cmd_sms_set_number(*args)
+        if args[0].length == 10
```
zeroSteiner: this should be ```if args[0].to_s.length == 10``` so if args[0] is nil because args.length == 0 it will still work.

data/msf/sms.rb
```diff
@@ -0,0 +1,186 @@
...
+        if args[0].length == 10 
+          print_status("Setting SMS number to #{args[0]}")
+          @sms_number = args[0]
+        else
+          print_error("Please provide a valid SMS number!")
+        end
+      end
+
+      def cmd_sms_set_carrier(*args)
+        if args.length > 0 and ['AT&T', 'Boost', 'Sprint', 'T-Mobile', 'Verizon', 'Virgin Mobile'].include? "#{args[0]}"
```
zeroSteiner: All the carriers should be a module constant, defined towards the top.

zeroSteiner:
All my feed back is pretty minor stuff and I was able to test this and it's working great. My only question is regarding the MIT license at the top, (everything else in King Phisher is BSD 3-clause) would you mind either updating it or giving me permission to change it for you? There's details on the differences in the "Comparison to other licenses" section of the [Wikipedia page](https://en.wikipedia.org/wiki/MIT_License).

There's a template for the BSD license in the [LICENSE](https://github.com/securestate/king-phisher/blob/master/LICENSE) where you could just update the top line.

zeroSteiner:
Landed this after some minor ruby changes in 514cc8789aa21254a44f11f2d6d4811d31b9e2c0, great work and thanks @coldfusion39!

PR #44, opened 2015-10-20
Support Phishing With Calendar Invites

zeroSteiner:
This semi-massive PR adds support for sending phishing messages in the form of calendar invites. To support this a new `king_phisher.ics` module was added and significant changes were made to the message configuration tab. All visible features are supported and fully functional. This has been tested with Microsoft Outlook both the web access and desktop application as well as with GMail. A known limitation in Google's Inbox service seems to be preventing calendar invites from working right now. Additionally a new `json_ex` module was added with logic taken from AdvancedHTTPServer to support JSON serialization of additional python object types, specifically `datetime.date` instances for the calendar widget.

Test steps:
 - [x] Normal message should continue to send correctly
 - [x] Calendar invite messages should work correctly
  - [x] Calendar invite message should support embedded images and attachments
 - [x] All of the unit tests should pass (check with `python -m unittest -v tests`)
 - [x] The documentation should build with no errors and make sense
 - [x] The GUI tooltips should clarify any questions regarding fields that must be filled out
 - [x] Exporting a message configuration to a KPM file with an invite should work correctly
  - [x] The same exported KPM file should also import correctly


wolfthefallen:
@zeroSteiner GUI tooltip is missing from "Calendar Invite Settings"; should read something to "Setting specific to sending calendar invite messages" similar to "Email Settings"
no tooltip on calendar invite drop down or radio buttons

PR #51, opened 2015-12-07
Added additonal filtering to the campaign selection diaglog

wolfthefallen:
Added gtk.menubutton with gtk.chkmenuitem for additional campaign filtering options. Removed gtk check box to filter expired campaigns, updated information about number of campaigns loaded into tree

* [x] Pass all unit testing
* [x] Pass King-phishers pylint rcfile
* [x] Able to filter out expired campaigns
* [x] Able to filter out current user campaigns
* [x] Able to filter out other users campaigns
* [x] Stores users filter setting on exit

@zeroSteiner

### DIFF
--- a/data/king-phisher.ipynb
+++ b/data/king-phisher.ipynb
@@ -32,7 +32,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {
     "collapsed": false
    },
@@ -48,6 +48,7 @@
     "import yaml\n",
     "\n",
     "king_phisher_home = widgets.Text(value='/opt/king-phisher', width='350px')\n",
+    "king_phisher_server_config = widgets.Text(value='server_config.yml', width='350px')\n",
     "db_connection = widgets.Text(placeholder='optional', width='350px')\n",
     "setup_button = widgets.Button(description='Setup', width='550px')\n",
     "results = widgets.HTML()\n",
@@ -69,13 +70,14 @@
     "    if db_connection.value:\n",
     "        db_manager.init_database(db_connection.value)\n",
     "    else:\n",
-    "        with open(os.path.join(directory, 'server-config.yml'), 'r') as file_h:\n",
+    "        with open(os.path.join(directory, king_phisher_server_config.value), 'r') as file_h:\n",
     "            server_config = yaml.load(file_h)\n",
     "        db_manager.init_database(server_config['server']['database'])\n",
     "    results.value = 'Successfully Initialized The Database'\n",
     "\n",
     "setup_button.on_click(setup_and_initialize)\n",
     "display.display(widgets.HBox([widgets.HTML(value='Install Directory', width='200px'), king_phisher_home]))\n",
+    "display.display(widgets.HBox([widgets.HTML(value='Server Configuration File', width='200px'), king_phisher_server_config]))\n",
     "display.display(widgets.HBox([widgets.HTML(value='Database Connection String', width='200px'), db_connection]))\n",
     "display.display(setup_button)\n",
     "display.display(results)"
@@ -181,6 +183,15 @@
     "session.commit()\n",
     "session.close()"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": []
   }
  ],
  "metadata": {


data/client/king_phisher/client_config.json
```diff
@@ -5,6 +5,9 @@
   "mailer.max_messages_per_connection": 5,
   "mailer.message_type": "email",
   "mailer.target_type": "file",
+  "menu_item_expired": false,
```
zeroSteiner: I'd like to see all these names changed and with a meaningful prefix. After this PR lands "menu_item_expired" will be a bit difficult to remember. Something like "filter.campaign.expired", "filter.campaign.other_users", etc. I think would be easier to remember.

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -57,13 +57,14 @@ class CampaignSelectionDialog(gui_utilities.GladeGObject):
 	gobject_ids = (
 		'button_new_campaign',
 		'button_select',
-		'checkbutton_show_expired',
 		'drawingarea_color_key',
 		'label_campaign_info',
-		'treeview_campaigns'
+		'treeview_campaigns',
+		'menubutton_filter'
```
zeroSteiner: This should be alphabetized with the rest.

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -57,13 +57,14 @@ class CampaignSelectionDialog(gui_utilities.GladeGObject):
...
 	)
+
 	top_gobject = 'dialog'
-	top_level_dependencies = ('StockAddImage',)
+	top_level_dependencies = ('StockAddImage')
```
zeroSteiner: This actually changes the meaning, in Python `('StockAddImage') != ('StockAddImage',)`. The first evaluates to a string, while the second (due to the trailing comma) evaluates to a tuple containing a single string object.

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -79,18 +80,70 @@ def __init__(self, *args, **kwargs):
 		self._tv_model.set_sort_column_id(5, Gtk.SortType.DESCENDING)
 		# create and set the filter for expired campaigns
 		self._tv_model_filter = self._tv_model.filter_new()
-		self._tv_model_filter.set_visible_func(self._filter_expired)
+		self._tv_model_filter.set_visible_func(self._filter_campaigns)
 		treeview.set_model(Gtk.TreeModelSort(model=self._tv_model_filter))
+
+		#setup menus for filtering campaigns and load campaigns
```
zeroSteiner: Single space after the comment for non-code comments.

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -79,18 +80,70 @@ def __init__(self, *args, **kwargs):
...
+		for menus in self.filter_menu_items:
+			filter_menu.append(self.filter_menu_items[menus])
+			self.filter_menu_items[menus].connect('toggled', self.signal_checkbutton_toggled)
+			self.filter_menu_items[menus].show()
+
+		try:
+			self.filter_menu_items['expired_campaigns'].set_active(self.config['menu_item_expired'])
+			self.filter_menu_items['your_campaigns'].set_active(self.config['menu_item_user'])
+			self.filter_menu_items['other_campaigns'].set_active(self.config['menu_item_other'])
+		except RuntimeError:
```
zeroSteiner: Under what conditions would this occur? If the config value doesn't exist wouldn't it throw a `KeyError`? Also because they're in the default config they should be added automatically.

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -79,18 +80,70 @@ def __init__(self, *args, **kwargs):
...
+
+		self.gobjects['menubutton_filter'].set_popup(filter_menu)
+		filter_menu.connect('destroy', self._save_filter)
+
+	def _save_filter(self, _):
+		self.config['menu_item_expired'] = self.filter_menu_items['expired_campaigns'].get_active()
+		self.config['menu_item_user'] = self.filter_menu_items['your_campaigns'].get_active()
+		self.config['menu_item_other'] = self.filter_menu_items['other_campaigns'].get_active()
+
+	def _filter_campaigns(self, model, tree_iter, _):
```
zeroSteiner: This function seems to contain a lot of duplicate logic. I think there's room for two improvements.
 1. Set `self.config['server_username']` to a var so the lines are shorter
 2. Consolidate the logic checks to one line or block each
```
if self.filter_menu_items['expired_campaigns'].get_active():
    if expiration_ts and utilities.parse_datetime(expiration_ts) > datetime.datetime.now():
        return True

if self.filter_menu_items['your_campaigns'].get_active() and campaign_owner != current_user:
    return True
...etc.
```

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -154,7 +207,7 @@ def load_campaigns(self):
 				(hlfg_color if is_expired else fg_color),
 				(html.escape(campaign.description, quote=True) if campaign.description else None)
 			))
-		self.gobjects['label_campaign_info'].set_text("Showing {0:,} Campaign{1}".format(len(store), ('' if len(store) == 1 else 's')))
+		self.gobjects['label_campaign_info'].set_text("{0:,} campaign{1} found in the database".format(len(store), ('' if len(store) == 1 else 's')))
```
zeroSteiner: Is there a way to count the number of campaigns that are being shown? If so this label should be kept the same because I think it would be useful for users to see how their filters are affecting the numbers of campaigns shown. For example if I filter out other users campaigns and it drops by 20 I can tell there are 20 campaigns from other users etc.

data/client/king_phisher/king-phisher-client.ui
```diff
@@ -3219,25 +3182,65 @@ Brandan Geise</property>
...
-                <property name="label" translatable="yes">Show Expired Campaigns</property>
+              <object class="GtkMenuButton" id="CampaignSelectionDialog.menubutton_filter">
                 <property name="visible">True</property>
                 <property name="can_focus">True</property>
-                <property name="receives_default">False</property>
-                <property name="xalign">0</property>
-                <property name="draw_indicator">True</property>
-                <signal name="toggled" handler="signal_checkbutton_toggled" swapped="no"/>
+                <property name="receives_default">True</property>
+                <property name="tooltip_text" translatable="yes">filter campaign options</property>
```
zeroSteiner: Tooltips should start with a capital and end with a period.

king_phisher/client/dialogs/campaign_selection.py
zeroSteiner: Also if this has to be changed then it should be updated in the glade file for consistency.

wolfthefallen:
@zeroSteiner updated!

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -79,19 +80,56 @@ def __init__(self, *args, **kwargs):
...
+		self.config['filter.campaign.other_users'] = self.filter_menu_items['other_campaigns'].get_active()
+
+	def _filter_campaigns(self, model, tree_iter, _):
 		expiration_ts = model[tree_iter][6]
-		if not expiration_ts:
-			return True
-		if utilities.parse_datetime(expiration_ts) > datetime.datetime.now():
+		campaign_owner = model[tree_iter][4]
+		username = self.config['server_username']
+		if self.filter_menu_items['your_campaigns'].get_active() == False:
```
zeroSteiner: `if not self.filter_menu_items['your_campaigns'].get_active():`

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -177,6 +219,11 @@ def signal_button_clicked(self, button):
 
 	def signal_checkbutton_toggled(self, _):
 		self._tv_model_filter.refilter()
+		self.gobjects['label_campaign_info'].set_text("Showing {0} of {1:,} campaign{2} found in the database.".format(
```
zeroSteiner: Modify the string to remove "found in database." and capitalize the C in campaign.

king_phisher/client/dialogs/campaign_selection.py
```diff
@@ -154,7 +192,11 @@ def load_campaigns(self):
 				(hlfg_color if is_expired else fg_color),
 				(html.escape(campaign.description, quote=True) if campaign.description else None)
 			))
-		self.gobjects['label_campaign_info'].set_text("Showing {0:,} Campaign{1}".format(len(store), ('' if len(store) == 1 else 's')))
+		self.gobjects['label_campaign_info'].set_text("Showing {0} of {1:,} campaign{2} found in the database.".format(
+																													len(self._tv_model_filter),
```
zeroSteiner: To be consistent with the other indentations of this style within the project, this line should be indented exactly one more than the line before it. The closing parenthesis should be aligned with the start of the line before this one.

wolfthefallen:
@zeroSteiner updated

zeroSteiner:
@wolfthefallen Looks good now! All tasks are done, I just made two very small changes and I'll land this shortly.

I did notice there was a change added that wasn't in the scope of the original PR for the Jupyter notebook but it looks alright as well so I don't see a need to remove it. Just in the future it should be a separate PR.

PR #58, opened 2015-12-17
updated install for fedora support

wolfthefallen:
@zeroSteiner Updated install.sh script to support Fedora 22 and 23, with utilizing dnf. Also checks for the missing folder to see if rpm-build needs to be installed.

- [ ] Successfully installs king-phisher on Fedora 22 and 23.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -65,7 +65,7 @@ if [ -z "$LINUX_VERSION" -a $? -eq 0 ]; then
 	KING_PHISHER_SKIP_CLIENT="x"
 fi
 
-grep -E "Fedora release 2[1-2]" /etc/redhat-release &> /dev/null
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -91,7 +91,7 @@ if [ -z "$LINUX_VERSION" ]; then
 	echo "  - BackBox"
 	echo "  - CentOS"
 	echo "  - Debian"
-	echo "  - Fedora"
+	echo "  - Fedora Release 22-23"
```
zeroSteiner: I've purposely avoided listing specific versions so that less code needs to be updated when the versions change. This should be consistent with the others and just say Fedora. The user should understand that the version is the latest and we should ensure that is accurate.

zeroSteiner:
Landed, thanks @wolfthefallen!

PR #67, opened 2016-02-09
Support An Alternative Python 2 ipaddress Module

zeroSteiner:
This adds support for using an alternative [ipaddress](https://pypi.python.org/pypi/ipaddress) module for Python 2.7 support. This wraps the module and normalizes the usage of the `address` paramter from a `str` to `unicode` to make it compatible with the currently used `py2-ipaddress` module.

**Notice:** To keep the package list clean, `py2-ipaddress` should be uninstalled with pip BEFORE `ipaddress` is installed in it's place.

### DIFF
--- a/KingPhisher
+++ b/KingPhisher
@@ -68,7 +68,7 @@ __requirements__ = [
 	'XlsxWriter>=0.7.7'
 ]
 if its.py_v2:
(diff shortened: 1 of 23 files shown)


wolfthefallen:
Landed! All unit tests pasts, after pip updating new requirements.

Issue #73, opened 2016-02-15
KingPhisher Client theme issue on backbox


wolfthefallen:
BackBox default theme is causing KingPhisher client cause the MailSenderConfigurationTab to appear almost completely white.

Current work around is to run KingPhisher client in no-style

```BASH
KingPhisher --no-style
```

piratica:
I was using an Ubuntu box for the KingPhisher client and don't currently have access to the BackBox server that we we setup.  I can check it tomorrow though and confirm that there are / are not styling issues and will be happy to report back.

zeroSteiner:
Just opened PR #74 to fix this issue. This should be closed out once it has been tested and landed.

zeroSteiner:
PR #74 has been landed to the dev branch so this issue has been fixed. The changes will eventually make their way into master within the next couple weeks.

PR #74, opened 2016-02-16
Fix Style For BackBox 4.5 / Gtk 3.10

zeroSteiner:
This PR fixes the style of GtkViewport and GtkNotebook tabs as reported in issue #73. This makes the mailer config tab visible and not all white in the client for BackBox 4.5. Additionally this fixes the color of the inactive GtkNotebook tabs. The GtkViewport issue was addressed by setting the background to be an image, this was the desired way to do it but resulted in an odd visual quark when the widget changed sizes due to the message type changing back and forth from calendar to email. This issue was addressed by subscribing to the `GtkExpander notify::expanded` signal and forcing the viewport to be redrawn.

Testing:
 - [x] Load the PR in BackBox 4.5 and make sure the GUI appears relatively normal.*

\* Relatively because of the differences between GTK 3.10 and newer it's not *exactly* the same.

### DIFF
--- a/data/client/king_phisher/king-phisher-client.ui
+++ b/data/client/king_phisher/king-phisher-client.ui
@@ -5682,9 +5682,6 @@ your communications. Please contact the system administrator.</property>
                     <property name="position">0</property>
                   </packing>
                 </child>
-                <style>
-                  <class name="background-add"/>
-                </style>
               </object>
               <packing>
                 <property name="expand">False</property>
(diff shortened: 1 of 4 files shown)


wolfthefallen:
Landed, This PR made the GUI beautiful in backbox 4.5, also fixed a few issues on Ubuntu 14.04 unity with gnome-themes-standard installed.

Issue #79, opened 2016-03-08
tools/install.sh fails on install of matplotlib from requirements.txt on Digital Ocean droplet


piratica:
## Issue Description
Attempting to setup a Digital Ocean droplet to test a site.  Tried via PTF and then directly but failed each time at matplotlib.  This seems to be specific to Digital Ocean droplets but seems to be distro independent.

## Reproduction Steps
Initially used PTF for the install but it failed.  Went directly into /pentest/exploitation/king-phisher and attempted to run tools/install.sh but failed again at the same spot.  Noted that pip had been installing matplotlib and had gotten to 99% and then bombed (I was too lazy to take them down but it was consistent across Ubuntu, Debian and Fedora droplets).

## Work Around so far
* install matplotlib via apt
  * sudo apt-get install python-mpltoolkits.basemap
  * sudo apt-get install python-mpltoolkits.basemap-data
* clean up apt (not sure why this is required yet but it seems to be)
  * sudo apt-get update --fix-missing
* comment the matplotlib in requirements.txt
  * sed -i 's/^matplotlib/#matplotlib/'
* tools/install.sh
* go phishing :)

zeroSteiner:
For Digital Ocean, the $20/month droplet needs to be used to have the 2GB of RAM in order to compile matplotlib. I ran into this error myself two weeks ago. You can temporarily either:
 * Upgrade to the 2GB level then downgrade when you're done
 * Mount disk space as extra swap space

When going through the install you'll notice an out of memory error as it's trying to decode some data.

piratica:
Very cool, thank you for the quick response on this.  I'll give that a shot and follow-up here but it seems like this is a non-issue.

piratica:
Ok, just spun up one of the $20 per month droplets and did the following:
* created droplet
* logged in via SSH
* apt-get install git
* git clone https://github.com/trustedsec/ptf.git
* (inside ptf) >use modules/explotation/kingphisher
* (inside ptf) >install
* (inside ptf) installer seemed to go well but bombed again at matplotlib
`Collecting matplotlib>=1.5.0 (from -r requirements.txt (line 13))
  Downloading matplotlib-1.5.1.tar.gz (54.0MB)
    100% |████████████████████████████████| 54.0MB 13kB/s 
    Complete output from command python setup.py egg_info:
    IMPORTANT WARNING:
        pkg-config is not installed.
        matplotlib may not be able to find some of its dependencies
    ============================================================================
[...]

PR #80, opened 2016-03-09
updated install script to handle issues with apt-get update package c…

wolfthefallen:
@zeroSteiner 

Update the tools/install.sh script to handle the issues identified in #79. apt-get update package cache was not update for installation, and sometimes fails when apt-get update occurs. Install script will now attempt to do apt-get update, and exit if it fails.

Example to run
```BASH
tools/install.sh
```
To test:
change source.list file so its wrong to force a 'fail'
- [ ]  script should exit on failing apt-get update

change source.list file back to normal.
- [ ] script should complete successfully

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -179,6 +179,15 @@ else
 	fi
 fi
 
+# update apt-get package information and only continue if successful
+if [ "$(command -v apt-get)" ]; then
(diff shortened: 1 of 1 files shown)


zeroSteiner:
Nice work @wolfthefallen

zeroSteiner:
Just landed this to the dev branch.

PR #99, opened 2016-04-13
Support for Ubuntu 16.04 LTS

wolfthefallen:
@zeroSteiner 
tools/install.sh script updated to support Ubuntu 16.04 LTS (Xenial Xerus)
tested on Ubuntu Xenial Xerus (development branch)

To test
- [x]  Git clone King Phisher
- [x]  cd to King Phisher and run sudo ./tools/install.sh

run install script for each option test case below, reverting VM in between installs.
- [x] successfully installs with postgresql option
- [ ] successfully installs with out postgresql option
- [ ] successfully install with client only

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -147,7 +147,7 @@ if [ -z "$LINUX_VERSION" -a $? -eq 0 ]; then
 	LINUX_VERSION="Kali"
 fi
 
-grep -E "Ubuntu 1[45]\.(04|10)" /etc/issue &> /dev/null
+grep -E "Ubuntu 1[456]\.(04|10)" /etc/issue &> /dev/null
 if [ -z "$LINUX_VERSION" -a $? -eq 0 ]; then
 	LINUX_VERSION="Ubuntu"
 fi
@@ -156,7 +156,6 @@ grep -E "Ubuntu Xenial Xerus" /etc/issue &> /dev/null
 if [ -z "$LINUX_VERSION" -a $? -eq 0 ]; then
 	LINUX_VERSION="Ubuntu"
 fi
-
 
 if [ -z "$LINUX_VERSION" ]; then
 	echo "Failed to detect the version of Linux"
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -263,10 +269,29 @@ elif [ "$LINUX_VERSION" == "BackBox" ] || \
...
+			exit
+		fi
+
+		apt-cache search gir1.2-vte-2.91
+		if [ $? -eq 0 ]; then
+			apt-get install gir1.2-vte-2.91
+		else
+			apt-get install gir1.2-vte-2.90
+		fi
+
```
zeroSteiner: This should check the exit status of the last command and complain that `gir1.2-vte-2.xx` couldn't be installed. I think it would be fine if the versions were combined for this check.

tools/install.sh
```diff
@@ -152,6 +152,12 @@ if [ -z "$LINUX_VERSION" -a $? -eq 0 ]; then
 	LINUX_VERSION="Ubuntu"
 fi
 
+grep -E "Ubuntu Xenial Xerus" /etc/issue &> /dev/null
+if [ -z "$LINUX_VERSION" -a $? -eq 0 ]; then
+	LINUX_VERSION="Ubuntu"
+fi
+
+
```
zeroSteiner: One blank line is sufficient here

tools/install.sh
```diff
@@ -263,10 +269,29 @@ elif [ "$LINUX_VERSION" == "BackBox" ] || \
 	 [ "$LINUX_VERSION" == "Ubuntu"  ]; then
 	apt-get install -y libfreetype6-dev python-dev python-pip
 	if [ -z "$KING_PHISHER_SKIP_CLIENT" ]; then
-		apt-get install -y gir1.2-gtk-3.0 gir1.2-gtksource-3.0 gir1.2-vte-2.90 \
+		apt-get install -y gir1.2-gtk-3.0 gir1.2-gtksource-3.0 \
+			gir1.2-webkit-3.0 python-cairo libgeos++-dev \
+			libgtk-3-dev libpq-dev python-gi python-gi-cairo libpq-dev \
+			python-gobject python-gobject-dev python-paramiko pkg-config
+		if [ $? -eq 0 ]; then
```
zeroSteiner: This logic should be inverted to check for the failure condition, print a simple error message and bail. The success logic should be the same level of indentation as this check. I also would like to avoid having all of the packages listed a second time in the error message. Telling the user an error occurred in this step would already be more than currently exists.

zeroSteiner:
Tested and landed, thanks @wolfthefallen. One less thing to worry about.

PR #106, opened 2016-04-25
Client Plugin Support

zeroSteiner:
This PR adds plugin support to the client and lays the ground work for plugins to be included in the server. There is not much of a "plugin API" yet, really there's only two convenience features supported by the `ClientPlugin` class:
 - Dedicated persistent configuration storage
 - Easy GTK signal connections with automatic removal when plugins are disabled

Client plugins are currently loaded from two locations:
 - `data/client/king_phisher/plugins` For plugins distributed by us as part of the project
 - `~/.config/king-phisher/plugins` For user-specific plugins

Plugins have a couple of states:
 - **Unloaded** The plugin exists on disk but has not been imported by Python into the process
 - **Loaded** The plugin has been imported by Python so the metadata can be accessed from the `Plugin` class
 - **Enabled** The plugin class has been initialized and is doing things

Right now the client "loads" all plugins on start up to allow the metadata to be accessed.

Testing:
 - [x] Review the basic API in `king_phisher/plugins.py` makes sense (it'll be used be the server too eventually)
 - [x] Build and read the docs
 - [x] Test the plugin manager in the GUI `Tools > Plugin Manager`
  - [x] Enabling plugins should work
  - [x] Disabling plugins should work
  - [x] Enabled plugins should persist across application runs
  - [x] Plugins meta data should be displayed in the manager
 - [x] The `--no-plugins` option disables all plugins on start up (to avoid any potential issues)

Here is a Hello World plugin suitable for testing. Once this is landed it will be included in the docs, pending any API changes that this PR review brings up.
```python
import king_phisher.client.plugins as plugins
import king_phisher.client.gui_utilities as gui_utilities

class Plugin(plugins.ClientPlugin):
	authors = ['Spencer McIntyre']
	title = 'Hello World!'
	description = 'A test plugin.'
	homepage = 'https://github.com/securestate/king-phisher'
	def initialize(self):
		print('Hello World!')
		self.signal_connect('exit', self.signal_exit)

	def finalize(self):
		print('Good Bye World!')

	def signal_exit(self, app):
		gui_utilities.show_dialog_info('Exiting!', app.get_active_window())
```

### DIFF
--- a/data/client/king_phisher/king-phisher-client.ui
+++ b/data/client/king_phisher/king-phisher-client.ui
@@ -2421,7 +2421,7 @@ Spencer McIntyre</property>
   </object>
   <object class="GtkApplicationWindow" id="PluginManagerWindow">
     <property name="width_request">550</property>
-    <property name="height_request">250</property>
+    <property name="height_request">300</property>
     <property name="can_focus">False</property>
     <property name="border_width">5</property>
     <property name="title" translatable="yes">Plugin Manager</property>
@@ -2462,7 +2462,7 @@ Spencer McIntyre</property>
             <property name="can_focus">True</property>
             <child>
               <object class="GtkScrolledWindow" id="PluginManagerWindow.scrolledwindow_plugin_info">
-                <property name="height_request">100</property>
+                <property name="height_request">125</property>
                 <property name="visible">True</property>
                 <property name="can_focus">True</property>
                 <property name="border_width">3</property>
@@ -2509,7 +2509,7 @@ Spencer McIntyre</property>
                           </object>
                           <packing>
                             <property name="left_attach">0</property>
-                            <property name="top_attach">1</property>
+                            <property name="top_attach">2</property>
                           </packing>
                         </child>
                         <child>
@@ -2526,7 +2526,7 @@ Spencer McIntyre</property>
                           </object>
                           <packing>
                             <property name="left_attach">2</property>
-                            <property name="top_attach">1</property>
+                            <property name="top_attach">2</property>
                           </packing>
                         </child>
                         <child>
@@ -2542,7 +2542,7 @@ Spencer McIntyre</property>
                           </object>
                           <packing>
                             <property name="left_attach">1</property>
-                            <property name="top_attach">1</property>
+                            <property name="top_attach">2</property>
                           </packing>
                         </child>
                         <child>
@@ -2558,7 +2558,7 @@ Spencer McIntyre</property>
                           </object>
                           <packing>
                             <property name="left_attach">3</property>
-                            <property name="top_attach">1</property>
+                            <property name="top_attach">2</property>
                           </packing>
                         </child>
                         <child>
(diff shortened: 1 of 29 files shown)


zeroSteiner:
This is turning into a beast of a PR. Support for defining options is now in as well.
Option types include:
 - String
 - Integer
 - Boolean
 - Enum (choice of strings)

The next commit (probably coming this weekend) will add documentation and probably additional option types building on those fundamental ones.
Additional option types I have in mind include:
 - IntegerRange
 - Port
 - Password
 - FilePath

zeroSteiner:
Plugins which utilize this functionality are now available in the new [king-phisher-plugins](https://github.com/securestate/king-phisher-plugins) repository. They'll be useful for testing.

data/client/king_phisher/king-phisher-client.ui
```diff
@@ -2418,6 +2419,251 @@ Spencer McIntyre</property>
...
+  <object class="GtkApplicationWindow" id="PluginManagerWindow">
+    <property name="width_request">550</property>
+    <property name="height_request">300</property>
+    <property name="can_focus">False</property>
+    <property name="border_width">5</property>
+    <property name="title" translatable="yes">Plugin Manager</property>
+    <property name="window_position">center-on-parent</property>
+    <property name="destroy_with_parent">True</property>
+    <child>
+      <object class="GtkBox" id="box29">
```
wolfthefallen: As King Phisher progresses in the build, we should start naming to identify the sub-components of the windows.` id = "PluginManangerWindow.main_box"` . Just in case we need to manipulate or call the object in a future.

data/client/king_phisher/king-phisher-client.ui
```diff
@@ -2418,6 +2419,251 @@ Spencer McIntyre</property>
...
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="orientation">vertical</property>
+        <child>
+          <object class="GtkScrolledWindow" id="scrolledwindow6">
```
wolfthefallen: same as above

data/client/king_phisher/king-phisher-client.ui
```diff
@@ -2829,11 +3075,11 @@ Spencer McIntyre</property>
           </packing>
         </child>
         <child>
-          <object class="GtkNotebook" id="ConfigurationDialog.notebook1">
+          <object class="GtkNotebook" id="ConfigurationDialog.notebook_config">
             <property name="visible">True</property>
             <property name="can_focus">True</property>
             <child>
-              <object class="GtkBox" id="box2">
+              <object class="GtkBox" id="ConfigurationDialog.box_server_tab">
```
wolfthefallen: like this woot!

zeroSteiner:
@wolfthefallen comments addressed, widgets named correctly.

wolfthefallen:
@zeroSteiner Thank you, this is an awesome addition to King Phisher and makes it very easy to add optional features to the project.

PR #109, opened 2016-05-05
Stop Loading Campaign Data When It Changes

zeroSteiner:
This changes the behavior of the loader threads responsible for loading the data seen in the Campaign tab. When the campaign-set signal is emitted, each of these threads spin up to refresh their respective views. Prior to this change, if the thread was busy while this signal was emitted, the load operation would simply be skipped. This changes that behavior to using a new `threading.Event` to signal that the thread loader should stop what it's doing. This is then used to stop threads which may be loading old campaign data to allow them to be restarted to load the new campaigns data. This issue can be encountered when opening a large campaign (think 10k+ messages) and then changing the campaign while the previous campaigns data was still being loaded.

Verification steps:
 - [x] Load a large campaign
 - [x] Immediately load a smaller campaign (while data from the first is still being loaded)
 - [x] See the data from the large campaign *immediately* stop loading as the new data starts
 - [x] Ensure campaigns can be changed to and from as would normally be expected

CC: @guitarmanj

### DIFF
--- a/king_phisher/client/tabs/campaign.py
+++ b/king_phisher/client/tabs/campaign.py
@@ -75,10 +75,33 @@ class CampaignViewGenericTab(gui_utilities.GladeGObject):
 		"""The thread object which loads the data from the server."""
 		self.loader_thread_lock = threading.Lock()
 		"""The :py:class:`threading.Lock` object used for synchronization between the loader and main threads."""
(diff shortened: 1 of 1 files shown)


wolfthefallen:
@zeroSteiner Thank you

PR #111, opened 2016-05-09
GObject signals

wolfthefallen:
@zeroSteiner Added GObject Signals

- campaign-created
- campaign-deleted
- config-load
- credential-deleted
- message-deleted
- message-sent
- server-disconnected
- sftp-client-start
- visit-deleted

- [ ] King Phisher functions run as intended
- [x] You can subscribe to the new GObjects
- [x] Documentation builds and is accurate.

### DIFF
--- a/docs/source/client_gui/gobject_signals.rst
+++ b/docs/source/client_gui/gobject_signals.rst
@@ -18,22 +18,23 @@ events.
    campaign.
    Signal emission stage set to run first.
 
+   :Emission Stage: ``RUN_FIRST``
    :param str campaign_id: The ID of the campaign whose information was changed.
 
 .. py:function:: campaign-created(campaign_id)
 
    This signal is emitted after the user creates a new campaign id. Subscribers
    to this signal can use it to conduct an action after a new campaign id is created.
-   Signal emission stage set to run first.
 
+   :Emission Stage: ``RUN_FIRST``
    :param str campaign_id: The ID of the new campaign.
 
-.. py:function:: campaign-delete(campaign_id)
+.. py:function:: campaign-deleted(campaign_id)
 
    This signal is emitted when the user deletes a campaign. Subscribers
    to this signal can use it to conduct an action after the campaign is deleted.
-   Signal emission stage set to run first.
 
+   :Emission Stage: ``RUN_LAST``
    :param str campaign_id: The ID of the campaign.
 
 .. py:function:: campaign-set(campaign_id)
@@ -41,30 +42,34 @@ events.
    This signal is emitted when the user sets the current campaign. Subscribers
    to this signal can use it to update and refresh information for the current
    campaign.
-   Signal emission stage set to run
 
+   :Emission Stage: ``RUN_FIRST``
    :param str campaign_id: The ID of the campaign.
 
-.. py:function:: config-load()
+.. py:function:: config-load(load_defaults)
 
    This signal is emitted when the client configuration is loaded from disk. This
    loads all of the clients settings used within the GUI.
    Signal emission stage set to run last.
+
+   :Emission Stage: ``RUN_FIRST``
+   :param bool load_defaults: Load defaults true or false.
 
 .. py:function:: config-save()
 
    This signal is emitted when the client configuration is written to disk. This
    saves all of the settings used within the GUI so they can be restored at a
    later point in time.
-   Signal emission stage set to run last.
 
-.. py:function:: credential-deleted(row_ids)
+   :Emission Stage: ``RUN_LAST``
+
+.. py:function:: credential-delete(row_ids)
 
    This signal is emitted when the user deletes a credential entry. Subscribers
    to this signal can use it to conduct an action an entry is deleted.
-   Signal emission stage set to run last.
 
-   :param str row_id: The row IDs that was deleted.
+   :Emission Stage: ``RUN_LAST``
+   :param list row_ids: The row IDs that are to be deleted.
 
 .. py:function:: exit()
 
@@ -73,27 +78,31 @@ events.
    client is disconnected from the server. At this point the exit operation can
    not be cancelled.
 
+   :Emission Stage: ``RUN_LAST``
+
 .. py:function:: exit-confirm()
 
    This signal is emitted when the client has requested that the application
    exit. Subscribers to this signal can use it as a chance to display a warning
    dialog and cancel the operation.
 
-.. py:function:: message-deleted(row_ids)
+   :Emission Stage: ``RUN_LAST``
+
+.. py:function:: message-delete(row_ids)
 
    This signal is emitted when the user deletes a message entry. Subscribers
    to this signal can use it to conduct an action an entry is deleted.
-   Signal emission stage set to run last.
 
-   :param str row_id: The row ID that was deleted.
+   :Emission Stage: ``RUN_LAST``
+   :param list row_ids: The row IDs that are to be deleted.
 
 .. py:function:: message-sent(target_uid, target_email)
 
    This signal is emitted when the user sends a message. Subscribers
    to this signal can use it to conduct an action after the message is sent,
    and the information saved to the database.
-   Signal emission stage set to run first.
 
+   :Emission Stage: ``RUN_FIRST``
    :param str target_uid: Message uid that was sent.
    :param str target_email: Email address associated with the sent message.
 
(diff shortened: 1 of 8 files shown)


king_phisher/client/application.py
```diff
@@ -411,6 +428,9 @@ def theme_file(self):
 			return DISABLED
 		return find.find_data_file(os.path.join('style', self._theme_file))
 
+	def do_config_load(self):
+		self.load_config(load_defaults=True)
+
 	def load_config(self, load_defaults=False):
```
zeroSteiner: This should become the new `do_config_load` method. Existing references to `load_config` should either A) be updated to `.emit('config-load')` or B) `load_config` should simply emit the signal. I would prefer solution A unless there are a lot of occurrences where the config is explicitly loaded (I don't think there are).

king_phisher/client/assistants/campaign.py
```diff
@@ -293,6 +293,7 @@ def signal_assistant_apply(self, _):
 		else:
 			try:
 				cid = self.application.rpc('campaign/new', campaign_name, description=campaign_description)
+				self.application.emit('campaign-created', cid)
```
zeroSteiner: This should not be within the exception handler.

king_phisher/client/tabs/campaign.py
```diff
@@ -172,8 +172,23 @@ def _prompt_to_delete_row(self, treeview, _):
 			return
 		if len(row_ids) == 1:
 			self.rpc('db/table/delete', self.remote_table_name, row_ids[0])
+			if self.remote_table_name == 'messages':
```
zeroSteiner: These six lines could be condensed to
```
if self.remote_table_name in ('credentials', 'messages', 'visits'):
    self.application.emit(self.remote_table_name[:-1] + '-deleted', row_ids[0])
```

king_phisher/client/tabs/campaign.py
```diff
@@ -172,8 +172,23 @@ def _prompt_to_delete_row(self, treeview, _):
...
+				self.application.emit('message-deleted', row_ids[0])
+			elif self.remote_table_name == 'visits':
+				self.application.emit('visit-deleted', row_ids[0])
+			elif self.remote_table_name == 'visits':
+				self.application.emit('credential-deleted', row_ids[0])
 		else:
 			self.rpc('db/table/delete/multi', self.remote_table_name, row_ids)
+			if self.remote_table_name == 'messages':
```
zeroSteiner: Same logic here to drop the "s" to create the signal name.

king_phisher/client/application.py
```diff
@@ -101,14 +101,24 @@ class KingPhisherClientApplication(_Gtk_Application):
...
+		'campaign-created': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
+		'campaign-deleted': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
 		'campaign-set': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
+		'config-load': (GObject.SIGNAL_RUN_LAST, None, ()),
 		'config-save': (GObject.SIGNAL_RUN_LAST, None, ()),
+		'credential-deleted': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
 		'exit': (GObject.SIGNAL_RUN_LAST, None, ()),
 		'exit-confirm': (GObject.SIGNAL_RUN_LAST, None, ()),
+		'message-deleted': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
+		'message-sent': (GObject.SIGNAL_RUN_LAST, None, (str, str)),
```
zeroSteiner: Since this is a dummy signal used as a hook point and the handler doesn't exist / is just pass it should probably be `GObject.SIGNAL_RUN_FIRST`.

king_phisher/client/application.py
```diff
@@ -101,14 +101,24 @@ class KingPhisherClientApplication(_Gtk_Application):
...
 		'reload-css-style': (GObject.SIGNAL_RUN_FIRST, None, ()),
 		'rpc-cache-clear': (GObject.SIGNAL_RUN_FIRST, None, ()),
-		'server-connected': (GObject.SIGNAL_RUN_LAST, None, ())
+		'server-connected': (GObject.SIGNAL_RUN_LAST, None, ()),
+		'server-disconnected': (GObject.SIGNAL_RUN_FIRST, None, ()),
+		'sftp-client-start': (GObject.SIGNAL_RUN_FIRST, None, ()),
```
zeroSteiner: Shouldn't this be `GObject.SIGNAL_RUN_LAST` so if a plugin wants to start a custom sftp client, they can do it before the default handler kicks in and mark the signal as handled?

docs/source/client_gui/gobject_signals.rst
```diff
@@ -46,14 +79,57 @@ events.
...
 
    This signal is emitted when the client has connected to the King Phisher
    server. The default handler sets the initial campaign optionally prompting
    the user to select one if one has not already been selected.
+   Signal emission stage set to run last.
+
+.. py:function:: server-disconnected()
+
+   This signal is emitted when the client has disconnected from the King Phisher
+   server. The default handler sets the initial campaign optionally prompting
```
zeroSteiner: The second sentence in this doesn't sound accurate.

docs/source/client_gui/gobject_signals.rst
```diff
@@ -46,14 +79,57 @@ events.
...
+   the user to select one if one has not already been selected.
+   Signal emission stage set to run first.
+
+.. py:function:: sftp-client-start()
+
+   This signal is emitted when the client starts sftp transfer from within
```
zeroSteiner: "starts the sftp client" not transfer, we don't know when or if a transfer will actually be initiated.

king_phisher/client/application.py
```diff
@@ -245,6 +255,12 @@ def campaign_configure(self):
 
 		assistant.interact()
 
+	def do_campaign_created(self, campaign_id):
+		pass
+
+	def do_campaign_deleted(self, campaign_id):
+		pass
```
zeroSteiner: This could be updated to be an actual handler:
```
self.rpc('db/table/delete', 'campaigns', campaign_id)
```
Then the caller would only need to emit the signal.

king_phisher/client/application.py
```diff
@@ -102,21 +102,21 @@ class KingPhisherClientApplication(_Gtk_Application):
 	__gsignals__ = {
 		'campaign-changed': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
 		'campaign-created': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
-		'campaign-deleted': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
+		'campaign-deleted': (GObject.SIGNAL_RUN_LAST, None, (str,)),
```
zeroSteiner: Since these are run last now, and you can use `connect_after` all of these "deleted" past-tense names should be "delete".

king_phisher/client/application.py
```diff
@@ -101,14 +101,24 @@ class KingPhisherClientApplication(_Gtk_Application):
 	# pylint: disable=too-many-public-methods
 	__gsignals__ = {
 		'campaign-changed': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
+		'campaign-created': (GObject.SIGNAL_RUN_FIRST, None, (str,)),
+		'campaign-deleted': (GObject.SIGNAL_RUN_LAST, None, (str,)),
```
zeroSteiner: This should be campaign-delete as well like the others.

king_phisher/client/application.py
```diff
@@ -411,7 +437,7 @@ def theme_file(self):
 			return DISABLED
 		return find.find_data_file(os.path.join('style', self._theme_file))
 
-	def load_config(self, load_defaults=False):
+	def do_config_load(self, load_defaults=False):
```
zeroSteiner: Since this is a signal handler, `load_defaults` should no longer be a kwarg, it should be required.

docs/source/client_gui/gobject_signals.rst
```diff
@@ -16,44 +16,132 @@ events.
...
+   :param str campaign_id: The ID of the campaign.
+
+.. py:function:: config-load(load_defaults)
+
+   This signal is emitted when the client configuration is loaded from disk. This
+   loads all of the clients settings used within the GUI.
+   Signal emission stage set to run last.
+
+   :Emission Stage: ``RUN_FIRST``
+   :param bool load_defaults: Load defaults true or false.
```
zeroSteiner: true or false is implied by the fact that it's a boolean, so this should be more descriptive. Can simple copy the docs from `do_config_load` of "Load missing options from the template configuration file."

zeroSteiner:
@wolfthefallen thanks, I implemented the last minor changes I commented on and landed this.

PR #112, opened 2016-05-11
Gobject signals

wolfthefallen:
Updated message-delete, visit-delete, and credential-delete from str to object to allow for [str, ...] to be successfully passed. updated documentation.


zeroSteiner:
Tested and landed @wolfthefallen. Verified by deleting multiple rows at once. Thanks!

PR #115, opened 2016-05-18
De-Linting

Zanice:
Squelched/fixed messages returned by PyLint when running on the King Phisher repository (specifically, $REPO/king_phisher). Running through PyLint now, with the config file found at $REPO/data/pylintrc, there are no remaining messages.

Message IDs C0411 (wrong-import-order) and R0204 (redefined-variable-type) were added to the config file under the "disable" blacklist as these can be ignored. The remaining messages thereafter were either squelched by case or fixed to resolve the message. The project should still be fully functional.


zeroSteiner:
These should all be addressed in the central [.pylintrc](https://github.com/securestate/king-phisher/blob/master/data/pylintrc) file instead of having disable statements added to each of the files.

Ideally there should be very very few inline exceptions where the rule is turned off on case-by-case basis which appears to have been the method of choice here. All in all only ~10% of the `# pylint: ` statements should be kept, the rest that are consistently repeated should be moved to the pylintrc file.

Zanice:
In attempting that with the codes I wanted to add to the config file, none of the additions took effect, and I confirmed I was using the right config file. I'll give it another go though and see if I can get that to work, as that indeed would be a much better solution.

zeroSteiner:
Were you specifying it with the `--rcfile=` option? Personally I have a symlink in the root directory for `.pylintrc -> data/pylintrc` so that it doesn't need to be specified but either solution *should* work.

Zanice:
I was, yes. I got the config file recognized by PyLint, just that none of my changes within the file seemed to take effect.

Zanice:
I updated the request, see the changes under the squashed commit.

zeroSteiner:
I'm running into some issues on my end with this. **The problem is not with the changes but rather with upgrading to the latest PyLint**. I keep getting a stack trace towards the end of the run complaining about a GI namespace.

```
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1191, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1161, in _load_backward_compatible
  File "/usr/lib64/python3.4/site-packages/gi/importer.py", line 114, in load_module
    introspection_module = get_introspection_module(namespace)
  File "/usr/lib64/python3.4/site-packages/gi/module.py", line 271, in get_introspection_module
    module = IntrospectionModule(namespace, version)
  File "/usr/lib64/python3.4/site-packages/gi/module.py", line 120, in __init__
    repository.require(namespace, version)
gi.RepositoryError: Requiring namespace 'JavaScriptCore' version '3.0', but '4.0' is already loaded
```

I'm going to mark this as delayed until I can figure out how to address this issue.

zeroSteiner:
There's an issue in the upstream pylint / astroid projects that is causing a failure when both `WebKit` and `WebKit2` are available.

PR #139, opened 2016-07-27
created uninstall.sh script

wolfthefallen:
Created uninstall script for King Phisher.

This script will remove all service files, .desktop, and icons. It has the option to remove the PostgreSQL database and King Phisher directory.

- [x] tools/uninstall --help
- [x] tools/uninstall | only removes service files for upstart and systemd
- [x] tools/uninstall --delete-database | does the base and warns of loss of campaign data.
- [x] tools/uninstall --delete-directory | does the base and warns that the King Phisher directory will be removed
- [x] tools/uninstall --delete-database --delete-directory -y | removes all the things

### DIFF
--- a/tools/uninstall.sh
+++ b/tools/uninstall.sh
@@ -1,0 +1,183 @@
+#!/bin/bash
+# vim: tabstop=4 softtabstop=4 shiftwidth=4 noexpandtab
+###############################################################################
+# This is the Linux installation script for the King Phisher Client and
+# Server on supported distributions.
+#
+# Project Home Page: https://github.com/securestate/king-phisher/
+# Authors:
+#   Erik Daguerre
+#
+# Supported Linux Distributions:
+#   Linux Flavor  | Client | Server |
+#   --------------|--------|--------|
+#   BackBox		  |   yes  |   yes  |
+#   CentOS		  |   no   |   yes  |
+#   Debian		  |   yes  |   yes  |
+#   Fedora		  |   yes  |   yes  |
+#   Kali		  |   yes  |   yes  |
+#   Ubuntu		  |   yes  |   yes  |
+#
+###############################################################################
+
+E_USAGE=64
+E_SOFTWARE=70
+E_NOTROOT=87
+FILE_NAME="$(dirname $(readlink -e $0) 2>/dev/null)/$(basename $0)"
+GIT_CLONE_URL="https://github.com/securestate/king-phisher.git"
+if [ -z "$KING_PHISHER_DIR" ]; then
+	KING_PHISHER_DIR="/opt/king-phisher"
+fi
+LINUX_VERSION=""
+
+answer_all_no=false
+answer_all_yes=false
+
+function prompt_yes_or_no {
+	# prompt the user to answer a yes or no question, defaulting to yes if no
+	# response is entered
+	local __prompt_text=$1
+	local __result_var=$2
+	if [ "$answer_all_no" == "true" ]; then
+		$__result_var="no";
+		return 0;
+	elif [ "$answer_all_yes" == "true" ]; then
+		eval $__result_var="yes";
+		return 0;
+	fi
+	while true; do
+		read -p "$__prompt_text [Y/n] " _response
+		case $_response in
+			"" ) eval $__result_var="yes"; break;;
+			[Yy]* ) eval $__result_var="yes"; break;;
+			[Nn]* ) eval $__result_var="no";  break;;
+			* ) echo "Please answer yes or no.";;
+		esac
+	done
+	return 0;
+}
+
+function show_help {
+	echo "Usage: $(basename $0) [-h] [-n/-y]"
+	echo ""
+	echo "King Phisher Install Script"
+	echo ""
+	echo "optional arguments"
+	echo "  -h, --help			show this help message and exit"
+	echo "  -n, --no			 answer no to all questions"
+	echo "  -y, --yes			 answer yes to all questions"
+	echo "  --delete-database	 deletes king-phisher database and all campaigns"
+	echo "  --delete-directory	removes king-phisher directory and all files"
+	return 0;
+}
+
+while :; do
+	case $1 in
+		-h|-\?|--help)
+			show_help
+			exit
+			;;
+		-n|--no)
+			if [ "$answer_all_yes" == "true" ]; then
+				echo "Can not use -n and -y together"
+				exit $E_USAGE
+			fi
+			answer_all_no=true
+			;;
+		-y|--yes)
+			if [ "$answer_all_no" == "true" ]; then
+				echo "Can not use -n and -y together"
+				exit $E_USAGE
+			fi
+			answer_all_yes=true
+			;;
+		--delete-database)
+			KING_PHISHER_DELETE_DATABASE="x"
+			;;
+		--delete-dirctory)
+			KING_PHISHER_DELETE_DIRECTORY="x"
+			;;
+		--)
+			shift
+			break
+			;;
+		-?*)
+			printf "Unknown option: %s\n" "$1" >&2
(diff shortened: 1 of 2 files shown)


tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+# Server on supported distributions.
+#
+# Project Home Page: https://github.com/securestate/king-phisher/
+# Authors:
+#   Erik Daguerre
+#
+# Supported Linux Distributions:
+#   Linux Flavor  | Client | Server |
+#   --------------|--------|--------|
+#   BackBox		  |   yes  |   yes  |
```
zeroSteiner: This indentation is off

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+# vim: tabstop=4 softtabstop=4 shiftwidth=4 noexpandtab
+###############################################################################
+# This is the Linux installation script for the King Phisher Client and
+# Server on supported distributions.
+#
+# Project Home Page: https://github.com/securestate/king-phisher/
+# Authors:
+#   Erik Daguerre
+#
+# Supported Linux Distributions:
```
zeroSteiner: I don't even think this is necessary for the uninstallation script. I'd prefer to see it removed so it does not need to be updated when the install script is updated.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
+#!/bin/bash
+# vim: tabstop=4 softtabstop=4 shiftwidth=4 noexpandtab
+###############################################################################
+# This is the Linux installation script for the King Phisher Client and
```
zeroSteiner: uninstallation script.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+fi
+
+if [ ! -z $KING_PHISHER_DELETE_DATABASE ]; then
+	echo "WARNING: Database will be deleted || ALL CAMPAIGN DATA WILL BE LOST"
+	LAST_CHANCE=""
+	prompt_yes_or_no "Are you sure you want to continue" LAST_CHANCE
+	if [ $LAST_CHANCE == "yes" ]; then
+		su postgres -c "psql -c \"DROP DATABASE king_phisher;\"" &> /dev/null
+		su postgres -c "psql -c \"DROP USER king_phisher;\"" &> /dev/null
+		echo "database removed"
```
zeroSteiner: `echo "The King Phisher database has been removed"`

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+	if [ $LAST_CHANCE == "yes" ]; then
+		if git status &> /dev/null; then
+			KING_PHISHER_DIR="$(git rev-parse --show-toplevel)"
+			echo "Git repo found at $KING_PHISHER_DIR"
+		elif [ -d "$(dirname $(dirname $FILE_NAME))/king_phisher" ]; then
+			KING_PHISHER_DIR="$(dirname $(dirname $FILE_NAME))"
+			echo "Project directory found at $KING_PHISHER_DIR"
+		fi
+		rm -rf $KING_PHISHER_DIR
+		echo "directory removed"
```
zeroSteiner: Same thing here regarding a more comprehensive message.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+		su postgres -c "psql -c \"DROP DATABASE king_phisher;\"" &> /dev/null
+		su postgres -c "psql -c \"DROP USER king_phisher;\"" &> /dev/null
+		echo "database removed"
+	fi
+fi
+
+if [ ! -z $KING_PHISHER_DELETE_DIRECTORY ]; then
+	echo "WARNING: $KING_PHISHER_DIR directory will be removed"
+	LAST_CHANCE=""
+	prompt_yes_or_no "Are you sure you want to continue" LAST_CHANCE
```
zeroSteiner: Move the confirmation logic to the start of the script so once all actions are confirmed it will run unimpeded.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+	esac
+	shift
+done
+
+if [ "$(id -u)" != "0" ]; then
+	echo "This must be run as root"
+	exit $E_NOTROOT
+fi
+
+
```
zeroSteiner: Only one blank line is necessary here.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+			if [ "$answer_all_no" == "true" ]; then
+				echo "Can not use -n and -y together"
+				exit $E_USAGE
+			fi
+			answer_all_yes=true
+			;;
+		--delete-database)
+			KING_PHISHER_DELETE_DATABASE="x"
+			;;
+		--delete-dirctory)
```
zeroSteiner: "directory" is misspelled here.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+	return 0;
+}
+
+function show_help {
+	echo "Usage: $(basename $0) [-h] [-n/-y]"
+	echo ""
+	echo "King Phisher Install Script"
+	echo ""
+	echo "optional arguments"
+	echo "  -h, --help			show this help message and exit"
```
zeroSteiner: These options should all be spaces instead of hard tabs. That way it will always be consistent, that's also how the install.sh script has them defined, it looks like these got switched to hard tabs.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+fi
+
+if [ -f /etc/init/king-phisher.conf ]; then
+	service king-phisher stop
+	rm /etc/init/king-phisher.conf
+fi
+
+if [ -f "/usr/local/share/applications/king-phisher.desktop" ]; then
+	rm /usr/local/share/applications/king-phisher.desktop
+	echo "removed king-phisher.desktop"
```
zeroSteiner: This echo statement and the one below it should also start with a capital.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+
+if [ -f "/usr/share/icons/hicolor/index.theme" -a "$(command -v gtk-update-icon-cache)" ]; then
+	echo "Updating the GTK icon cache"
+	gtk-update-icon-cache --force /usr/share/icons/hicolor
+fi
+
+if [ $answer_all_yes == "true" ]; then
+	KING_PHISHER_DELETE_DATABASE="x"
+	KING_PHISHER_DELETE_DIRECTORY="x"
+	answer_all_yes=false # reset to allow for last chance
```
zeroSteiner: The last chance should be checked here and the variable shouldn't be reset.

zeroSteiner: Also if the user specifies yes for all then there probably shouldn't be a confirmation. Otherwise the option is misleading.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+#   Kali		  |   yes  |   yes  |
+#   Ubuntu		  |   yes  |   yes  |
+#
+###############################################################################
+
+E_USAGE=64
+E_SOFTWARE=70
+E_NOTROOT=87
+FILE_NAME="$(dirname $(readlink -e $0) 2>/dev/null)/$(basename $0)"
+GIT_CLONE_URL="https://github.com/securestate/king-phisher.git"
```
zeroSteiner: I don't think this is used anywhere.

tools/uninstall.sh
```diff
@@ -0,0 +1,183 @@
...
+if [ -z "$KING_PHISHER_DIR" ]; then
```
zeroSteiner: This and the 2 lines below it don't seem necessary, the directory can be found using the two methods you have implemented towards the end.

PR #155, opened 2016-10-05
removed sms sender so the message sends to a t-mobile phone

guitarmanj:
@zeroSteiner @wolfthefallen


zeroSteiner:
I've got this @wolfthefallen, we tested this in the office a while back so I know it's good to land.

zeroSteiner:
Landed, thanks @guitarmanj!

PR #173, opened 2016-11-24
Close the websocket on RPC logout

zeroSteiner:
This PR changes how the client shutsdown the web socket. Instead of closing it on the client side and waiting for the event loop to exit out, the client lets the server shut it down as part of the logout process. This means that the connection close is initiated from the server side instead of the client. This seems to allow the `run_forever` loop to process the servers close message and exit correctly. There's an internal flag `reconnect` that just needs to be set to False letting the event socket know that we're done with it and it doesn't need to try and reconnect to the server.

The server should have been shutting down the event socket if it existed when the client logged out anyways.

- [x] Test that this fixes the periodic logout hang when the client goes to exit the application

There is another `ConnectionResetError` exception that sometimes get's raised. That's a separate issue that is less of a concern because the client can still exit instead of hanging indefinitely. That issue can be tracked with the KP-45 ticket. Disregard that for now, I'll fix it next.

### DIFF
--- a/king_phisher/client/application.py
+++ b/king_phisher/client/application.py
@@ -388,10 +388,6 @@ class KingPhisherClientApplication(_Gtk_Application):
 
 	def do_exit(self):
 		self.plugin_manager.shutdown()
-		if self.server_events is not None:
-			self.server_events.shutdown()
-			self.server_events = None
-
 		self.main_window.hide()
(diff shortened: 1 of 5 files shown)


wolfthefallen:
King Phisher client is now logging out cleanly with out hanging!

Issue #169, opened 2016-11-14
Can't see info of messages sent for a campaign if the target CSV contains UTF-8 characters.


twkrol:
## Issue Description
Can't see info of messages sent for a campaign if the recipients data in target CVS file contains UTF-8 characters.

## Reproduction Steps
1. Prepare target CSV file with non-standard UTF-8 characters (lets use polish chars like ąćęłśńóźć), for example (included https://github.com/securestate/king-phisher/files/589208/ldap-emails.testy.txt)
[ldap-emails testy.txt]:
"john.doe2","John Doe2","john.doe@acme.com","Marcom ąćęłńóśżź"
1. Start King Phisher client for Windows (v1.5.1)
2. Open campaign, select target CSV file above and send mailing
3. The mailing should go, you can see incremented Messages count at View Campaign/Dashboard but no listing of sent emails at View Campaign/Messages (at least listing stops at that newly added record with special chars)
4. King Phisher client crashes every few times you click Refresh for View Campaign (possibly related)

## Environment Details
Host OS: Ubuntu 16.04 LTS

## Error Details / Stack Trace
nothing interesting here

zeroSteiner:
I'm going to see if I can reproduce this. Whether I can or not though, I'd prefer to have a stack trace if there is a crash so I can ensure that I was able to reproduce the exact error you encountered. Would you please be able to post the details of the crash, there should be information in the exception window.

twkrol:
The problem with messaging view exists all the time. Crashing client happened on clicking Refresh many times.
And there is no error on client fail, but I tried with VS Debugging and got:

> Unhandled exception at 0x66840E64 (libgtk-3-0.dll) in KingPhisher.exe: 0xC0000005: Access violation reading location 0x00000034.

But no more usefull info.

zeroSteiner:
Ok that's still helpful so thanks for the additional information.

twkrol:
KingPhisher client failing call stack reported by VS is:

> >	libgtk-3-0.dll!66840e64()	Unknown
 	[Frames below may be incorrect and/or missing, no symbols loaded for libgtk-3-0.dll]	
 	libgtk-3-0.dll!66843ce9()	Unknown
 	libglib-2.0-0.dll!687e2a1d()	Unknown
 	libglib-2.0-0.dll!687e298d()	Unknown
 	libgtk-3-0.dll!6685498c()	Unknown
 	libgdk-3-0.dll!710c1c2f()	Unknown
[...]

PR #179, opened 2017-01-10
Corrected unicode and utf8 issue

wolfthefallen:
This is to correct the issue #169. As information is being loaded into the GTK-TreeView(s) it will check to see if that cell of information is in `bytes` and if so it will decode it to `utf-8`. By doing this the windows application that is running python2 will now properly display special characters. 

To test, Follow instructions from #169 in a windows client, or running King Phisher in python2. Special characters such as `ąćęłńóśżź` should now correctly display in the department field.

Create a new campaign with special characters. It should correctly populate with out a stack trace.

- [x] Department field now displays special characters
- [ ] Campaign Names now can have special characters

### DIFF
--- a/king_phisher/client/application.py
+++ b/king_phisher/client/application.py
@@ -367,7 +367,7 @@ class KingPhisherClientApplication(_Gtk_Application):
 				)
 
 	def do_campaign_set(self, campaign_id):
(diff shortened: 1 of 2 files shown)


zeroSteiner:
I definitely see how this change fixes all of the string fields in the Campaign tab, but I don't see how it changes anything for how the Campaign Name is shown in the Campaign selection dialog (or anywhere else that the Campaign Name is used).

PR #188, opened 2017-02-01
Support Messages With To/CC/BCC

zeroSteiner:
This PR adds the ability for messages to be sent with arbitrary values in the To/CC/BCC fields. The King Phisher operator can now select a target field, this is the field that the target recipient is placed in. Additionally to that, the remaining headers can be adjusted to give the appearance that other users are included on the message when it is **not** delivered to them.

The default behavior mimics the old and that is to set Target Field to "To" and leave the "CC Email" field blank. Right now if the target field is not To, then the "To Email" field needs to be set.

- [x] Send an email with Target Field "To" and a made up CC Email address
  - [x] Ensure that the email arrives and the specified address shows up in the CC field
- [x] Send an email with the Target Field "CC" and a made up To Email Address
  - [x] Ensure that the email arrives and the specified address shows up in the To field
- [x] Send an email with the Target Field "BCC" and a made up To Email Address
  - [x] Ensure that the email arrives and the specified address shows up in the To field
- [x] Read all of the new tooltips and ensure that they make sense
- [x] Ensure that the settings persist across application runs

### DIFF
--- a/data/client/king_phisher/client_config.json
+++ b/data/client/king_phisher/client_config.json
@@ -4,6 +4,7 @@
   "dashboard.top_right": "VisitorInfo",
   "mailer.max_messages_per_connection": 5,
   "mailer.message_type": "email",
+  "mailer.target_field": "to",
   "mailer.target_type": "file",
   "filter.campaign.expired": false,
   "filter.campaign.user": true,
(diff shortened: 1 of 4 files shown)


wolfthefallen:
This is Awesome! Works like a charm. I love how you can select what field you want the target email address to be placed, and then populate the other fields with fake, real, other email address. When sending the phish. These addition email address will appear as being sent to the receipt but never truly sent. This will make for some fantastic phishing. 

Only thing that is really not mentioned is you can add more then one email address in the cc, bcc, or to fields to make it appear more then one person is receiving or copied on the email. This can be done by doing `email@email.com, email2@anydomain.net, ect@ect.com`.

PR #189, opened 2017-02-06
Added alembic_version table handling

wolfthefallen:
This pull request checks if the alembic_version table is present prior to trying to updating the database schema. If the table is not present it will create it and set it to the alembic version based off the version in the meta_data table.

Updated Based on comments.. alembic version table create will now occur if the table is not present, regardless if an schema update is required or not.

testing of alembic version table creation.
- [x] Verify alembic_version table is not present, if so drop it
- [x] Start King Phisher Server
- [x] Logging out put makes sense, and it sets alembic version
- [x] Verify alembic_version table was created and correct version is set.

bug testing
- [ ] git clone king-phisher repository
- [ ] `git checkout v1.5.2`
- [ ] `sudo tools/install.sh`
- [ ] start the king phisher server, and then stop it after it starts up completely
- [ ] very there is no alembic_version table
- [ ] `git checkout v1.6.0`
- [ ] 'sudo tools/install.sh'
- [ ] 'cp server_config.yml.bck server_config.yml`
- [ ] merge in this PR
- [ ] `./KingPhisherServer -L DEBUG server_config.yml -f`
- [ ] info logger stating alembic version was set to `a695de64338`
- [ ] schema upgrade completes successfully

### DIFF
--- a/king_phisher/server/database/manager.py
+++ b/king_phisher/server/database/manager.py
@@ -68,6 +68,14 @@ _flush_signal_map = (
 _meta_data_type_map = {'int': int, 'str': str}
 _popen = lambda args: subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
 
+ALEMBIC_VERSIONS = {
+	3: '24a4a626ff7c',
+	4: '7c315088952',
+	5: '83e4121b299',
+	6: 'a695de64338',
+	7: 'b76eab0a059'
+}
+
 @sqlalchemy.event.listens_for(Session, 'after_flush')
 def on_session_after_flush(session, flush_context):
 	for signal, session_attribute in _flush_signal_map:
@@ -280,6 +288,16 @@ def init_database(connection_url, extra_init=False):
 		alembic_directory = find.find_data_directory('alembic')
 		if not alembic_directory:
 			raise errors.KingPhisherDatabaseError('cannot find the alembic data directory')
+
+		if not 'alembic_version' in inspector.get_table_names():
+			alembic_metadata = sqlalchemy.MetaData(engine)
+			alembic_table = sqlalchemy.Table('alembic_version', alembic_metadata,
+								sqlalchemy.Column('version_num', sqlalchemy.String, primary_key=True, nullable=False)
+							)
+			alembic_metadata.create_all()
+			alembic_version_entry = alembic_table.insert().values(version_num=ALEMBIC_VERSIONS[schema_version])
+			engine.connect().execute(alembic_version_entry)
+			logger.info('alembic version set to {}'.format(ALEMBIC_VERSIONS[schema_version]))
 
 		config = alembic.config.Config(alembic_config_file)
 		config.config_file_name = alembic_config_file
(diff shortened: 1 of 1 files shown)


zeroSteiner:
This PR addresses a bug in how KP servers that do not have a database table called `alembic_version` fail to upgrade their schema correctly due to alembic not knowing the current revision.

Manual fix in the mean time:
```
cd data/server/king_phisher
alembic -x config=/opt/king-phisher/server_config.yml stamp a695de64338
alembic -x config=/opt/king-phisher/server_config.yml upgrade head
```

king_phisher/server/database/manager.py
```diff
@@ -281,6 +289,16 @@ def init_database(connection_url, extra_init=False):
 		if not alembic_directory:
 			raise errors.KingPhisherDatabaseError('cannot find the alembic data directory')
 
+		if not 'alembic_version' in inspector.get_table_names():
+			alembic_metadata = sqlalchemy.MetaData(engine)
+			alembic_table = sqlalchemy.Table('alembic_version', alembic_metadata,
```
zeroSteiner: The whitespace here is off and this block should probably all be put in a new `init_alembic` function.

king_phisher/server/database/manager.py
```diff
@@ -68,6 +68,14 @@
 _meta_data_type_map = {'int': int, 'str': str}
 _popen = lambda args: subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=True)
 
+ALEMBIC_VERSIONS = {
```
zeroSteiner: This information should be pulled dynamically by listing and parsing the versions directory. Something like:
```py
alembic_directory = find.find_data_directory('alembic')
versions = os.listdir(os.path.join(alembic_directory, 'versions'))
# parser versions into a dictionary possibly using a regex
```

zeroSteiner:
@wolfthefallen I added a unit test to check the schema file names adhere to the standard in commit 76d9d9e76afbed79a513f1b91f6bb8e0b6667ad5, so that part is done. You can find the matching regex in that commit, it should be simple enough to parse.

zeroSteiner:
I modified the testing steps to reproduce the bug a bit since I didn't want to reinstall King Phisher. What I did was used alembic to downgrade myself from v7 to v6, then deleted the alembic_version table and finally restart King Phisher. The relevant lines from the log are below and it worked as expected.

```
INFO     KingPhisher.Server.Database              initializing database connection with driver postgresql
DEBUG    KingPhisher.Server.Database              using postgresql-setup to ensure that the database is initialized
DEBUG    KingPhisher.Server.Database              postgresql service is already running via systemctl
DEBUG    KingPhisher.Server.Database              current database schema version: 6 (obsolete)
DEBUG    KingPhisher.Server.Database              alembic version table not found, attempting to create and set version
INFO     KingPhisher.Server.Database              alembic_version table initialized to a695de64338
WARNING  KingPhisher.Server.Database              automatically updating the database schema from version 6 to 7
INFO     alembic.runtime.migration                Context impl PostgresqlImpl.
INFO     alembic.runtime.migration                Will assume transactional DDL.
INFO     alembic.runtime.migration                Running upgrade a695de64338 -> b76eab0a059, Schema v7
DEBUG    alembic.runtime.migration                update a695de64338 to b76eab0a059
INFO     KingPhisher.Server.Database              successfully updated the database schema from version 6 to 7
DEBUG    KingPhisher.Server.Database              connected to postgresql database: king_phisher
INFO     KingPhisher.Server.SessionManager        restored 0 valid sessions and skipped 0 expired sessions from the database
```
I made some minor modifications and this will be landed shortly.

PR #192, opened 2017-02-20
RPC insert multi

wolfthefallen:
This PR will add RPC `/db/table/insert/multi` functionality to insert multiple rows in to a database. Additionally if `deconflict_ids` is true to will generate new ID to add that row into the database. To support this `make_uid` functionality for messages and visits have been moved to `utilities.py` to allow both client and server access to these functions. 

- [x] Function description for `/db/table/insert/multi` makes since

Testing:
- [x] Pull down PR and start King Phisher server and Client utilizing new library
- [x] Run a test campaign to verify message uids are being generated correctly
- [x] Visit pages from test campaign to verify visit uids are being generated corrrectly
- [x] Start RPC Terminal in King Phisher Client
- [x] use rpc.('/db/table/insert/multi', str(table_name), list(keys), list((values,), (values,))) to insert multiple rows with specified values into target table.
-  [x] use rpc.('/db/table/insert/multi', str(table_name), list(keys), list((values,), (values,)), deconflict_ids=True) to insert multiple rows with specified values into target table and and deconflict ids of rows.

### DIFF
--- a/king_phisher/server/server_rpc.py
+++ b/king_phisher/server/server_rpc.py
@@ -39,6 +39,7 @@ from king_phisher import errors
 from king_phisher import geoip
 from king_phisher import ipaddress
 from king_phisher import version
+from king_phisher import utilities
 from king_phisher.constants import ConnectionErrorReason
 from king_phisher.server import graphql
 from king_phisher.server import signals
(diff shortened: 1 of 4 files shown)


king_phisher/server/server_rpc.py
```diff
@@ -493,6 +494,63 @@ def rpc_database_insert_row(handler, session, table_name, keys, values):
...
+			row_id = row[keys.index('id')]
+			check_id = db_manager.get_row_by_id(session, database_table_objects[table_name], row_id)
+			while check_id is not None:
+				row = list(row)
+				row_id = row[keys.index('id')]
+				if table_name == 'visits':
+					row_id = utilities.make_visit_uid()
+				elif table_name == 'messages':
+					row_id = utilities.make_message_uid()
+				elif table_name in ('landing_pages', 'credentials'):
```
zeroSteiner: Tables that have their primary key as an integer should have their new row ID set to `None`, the database will automatically assign the next available value. Only primary keys that are strings need to be set manually, like you are already doing for `visits` and `messages`.

king_phisher/server/server_rpc.py
```diff
@@ -493,6 +494,63 @@ def rpc_database_insert_row(handler, session, table_name, keys, values):
...
+	table = database_table_objects.get(table_name)
+	if not table:
+		raise errors.KingPhisherAPIError('failed to get table object for: {0}'.format(table_name))
+	for key in keys:
+		if key not in database_tables[table_name]:
+			raise errors.KingPhisherAPIError('column {0} is invalid for table {1}'.format(keys, table_name))
+
+	for row in rows:
+		if len(row) != len(keys):
+			errors.KingPhisherAPIError('row is not the same as number of values defined')
```
zeroSteiner: `row is not the same length as the number of values defined`

king_phisher/server/server_rpc.py
```diff
@@ -493,6 +494,63 @@ def rpc_database_insert_row(handler, session, table_name, keys, values):
...
+		if deconflict_ids and 'id' in keys:
+			row_id = row[keys.index('id')]
+			check_id = db_manager.get_row_by_id(session, database_table_objects[table_name], row_id)
```
zeroSteiner: The `table` variable can be used here instead of `database_table_objects[table_name]`.

king_phisher/server/server_rpc.py
```diff
@@ -493,6 +494,63 @@ def rpc_database_insert_row(handler, session, table_name, keys, values):
...
+	if not table:
+		raise errors.KingPhisherAPIError('failed to get table object for: {0}'.format(table_name))
+	for key in keys:
+		if key not in database_tables[table_name]:
+			raise errors.KingPhisherAPIError('column {0} is invalid for table {1}'.format(keys, table_name))
+
+	for row in rows:
+		if len(row) != len(keys):
+			errors.KingPhisherAPIError('row is not the same as number of values defined')
+		if deconflict_ids and 'id' in keys:
```
zeroSteiner: Before entering this block, it might be easiest to convert row to a dictionary like: `row = dict(zip(keys, row))` so you can more easily check the `id` field exists, retrieve it's value etc. It would also allow you to later on replace the row object creation logic:

```
 +		table_row = table()
 +		for key, value in zip(keys, row):
 +			setattr(table_row, key, value)
```
with: `table_row = table(**row)`.

PR #196, opened 2017-03-21
XML campaign import

wolfthefallen:
# Overview
This PR adds a new feature to King Phisher to allow for a user to import campaigns from King Phisher XML files version 1.3 and higher.

# Steps to use
- [x] pull down PR
- [x] Start King Phisher Client and Login to server
- [x] Export a Campaign
- [x] From the file menu select `Import/XML Campaign`
- [x] A new window pops up
- [x] Click Open and navigate to the XML file of the campaign you exported.
  - [x] Import Campaign is insensitive, because the name of the campaign is already in use
- [x] Rename the campaign, by changing the name in the second text entry field
  - [x] Import Campaign button becomes sensitive on a campaign name that is not in the database
- [x] Click Import Campaign
- [x] Window provides feed back to user during import with out locking main thread
  - [x] Progress bar updates
  - [x] Progress of steps update in the expander's text field
  - [x] Expander's Spinner starts and runs during import
- [ ] Campaign successfully imports

additional checking
- [x] Exiting the Window during import prompts the user if they want to stop import.
  - [x] If yes, Thread cleanly exits, and deletes created campaign for import
- [x] During import the Open button is insensitive
- [x] Documents build and are accurate

### DIFF
--- a/docs/source/change_log.rst
+++ b/docs/source/change_log.rst
@@ -17,6 +17,7 @@ Version 1.7.0
 * New features for client and server plugins
 * Add comparison of "trained" statistics to the campaign comparison
 * Support for including and importing Jinja templates from relative paths
+* New feature to import Campaigns from XML files.
 
 Version 1.6.0
 ^^^^^^^^^^^^^
--- a/docs/source/change_log.rst
+++ b/docs/source/change_log.rst
@@ -17,7 +17,7 @@ Version 1.7.0
 * New features for client and server plugins
 * Add comparison of "trained" statistics to the campaign comparison
 * Support for including and importing Jinja templates from relative paths
-* New feature to import Campaigns from XML files.
+* New feature to import Campaigns from XML files
 
 Version 1.6.0
 ^^^^^^^^^^^^^
--- a/king_phisher/client/windows/campaign_import.py
+++ b/king_phisher/client/windows/campaign_import.py
@@ -35,7 +35,6 @@ import logging
 import threading
 import xml.etree.ElementTree as ET
 
-from king_phisher import utilities
 from king_phisher import serializers
 from king_phisher.client import gui_utilities
 from king_phisher.client.widget import extras
@@ -97,11 +96,18 @@ class ImportCampaignWindow(gui_utilities.GladeGObject):
 		self.thread_import_campaign = None
 
 	def _set_text_view(self, string_to_set):
-		self.text_buffer.set_text(string_to_set + "\n")
+		GLib.idle_add(self.text_buffer.set_text, string_to_set + "\n")
 
-	def _update_text_view(self, string_to_add):
+	def __update_text_view(self, string_to_add):
 		end_iter = self.text_buffer.get_end_iter()
 		self.text_buffer.insert(end_iter, string_to_add + "\n")
+
+	def _update_text_view(self, string_to_add, idle=False):
+		if idle:
+			GLib.idle_add(self.__update_text_view, string_to_add)
+		else:
+			end_iter = self.text_buffer.get_end_iter()
+			self.text_buffer.insert(end_iter, string_to_add + "\n")
 
 	def _update_id(self, element, id_fields, old_id, new_id):
 		"""
@@ -138,7 +144,7 @@ class ImportCampaignWindow(gui_utilities.GladeGObject):
 		self.window.set_position(Gtk.WindowPosition.CENTER_ON_PARENT)
 		self.window.show_all()
 
-	def entry_change(self, _):
+	def signal_entry_change(self, _):
 		"""
 		When there is a change in the campaign entry field it will check to see if the name is already in use.
 		If it is not in use it will change the Sensitivity of the Import Campaign Button to allow the user to
@@ -152,29 +158,30 @@ class ImportCampaignWindow(gui_utilities.GladeGObject):
 		self.button_import_campaign.set_sensitive(True)
 		return
 
-	def open_xml_file(self, _):
+	def signal_multi_open_xml_file(self, _):
 		if not self.select_button.get_property('sensitive'):
 			return
 		self.select_xml_campaign()
 
-	def close_import_campaign(self, _):
+	def signal_multi_close_import_campaign(self, _, __):
 		"""
 		Checks to make sure the import campaign thread is closed before closing.
 		"""
 		if not self.campaign_info:
-			return
+			return False
 		if not self.thread_import_campaign:
-			return
-		if not self.thread_import_campaign.isAlive():
-			return
+			return False
+		if not self.thread_import_campaign.is_alive():
+			return False
 		response = gui_utilities.show_dialog(
 			Gtk.MessageType.QUESTION,
 			'Do you want to cancel importing the campaign?',
 			self.window,
 			message_buttons=Gtk.ButtonsType.YES_NO
 		)
-		if not response:
-			return
+		print(response)
+		if not response or response == -9:
+			return True
 
 		self.thread_import_campaign.stop()
 		self.thread_import_campaign.join()
@@ -203,16 +210,16 @@ class ImportCampaignWindow(gui_utilities.GladeGObject):
 			self.rpc('db/table/delete', 'campaigns', campaign_id)
 			self.logger.info('Deleted Campaign {}'.format(campaign_id))
 
-	def import_button(self, _):
+	def signal_import_button(self, _):
 		"""
 		Function called when the user starts the import campaign process, It will create a new thread,
 		to have the campaign run in the background freeing up the GUI for the user to conduct other functions.
 		"""
 		if not self.campaign_info:
-			GLib.idle_add(self._update_text_view, 'No campaign information to import')
+			self._update_text_view('No campaign information to import')
 			self.button_import_campaign.set_sensitive(False)
 			return
-		self.thread_import_campaign = ImportThread(target=self.import_campaign)
+		self.thread_import_campaign = ImportThread(target=self._import_campaign)
 		self.thread_import_campaign.start()
 
 	def select_xml_campaign(self):
@@ -269,14 +276,11 @@ class ImportCampaignWindow(gui_utilities.GladeGObject):
 
 		if next((nodes for nodes in self.db_campaigns if nodes['node']['name'] == campaign_name), None):
 			if verbose:
-				GLib.idle_add(
-					self._update_text_view,
-					'Campaign name {} is already in use by another campaign.'.format(campaign_name)
-				)
+				self._update_text_view('Campaign name {} is already in use by another campaign.'.format(campaign_name), idle=True)
 			return False
 
 		if verbose:
-			GLib.idle_add(self._update_text_view, 'Campaign Name {} is not in use, ready to import'.format(campaign_name))
+			self._update_text_view('Campaign Name {} is not in use, ready to import'.format(campaign_name), idle=True)
 		return True
 
 	def prepep_xml_data(self):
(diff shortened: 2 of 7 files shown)


zeroSteiner:
Got an unhandled exception while testing:
```
Error Type: builtins.AttributeError
Error Details: AttributeError("'ImportCampaignWindow' object has no attribute 'select_button'",)
Error UID: e1fd7118-c269-48de-ab9b-1a2e9342e198
RPC Error: N/A (Not a remote RPC error)
King Phisher Version: 1.7.0-beta2 (rev: 746b2eaa777a)
Platform Version: Linux: Fedora 25 Twenty Five
Python Version: 3.5.3
Gtk Version: 3.22.10
Timezone: America/New_York

Thread Information:
    Thread-1 (alive=True daemon=True)

Stack Trace:
  File "/home/steiner/repos/king-phisher/king_phisher/client/windows/campaign_import.py", line 156, in open_xml_file
    if not self.select_button.get_property('sensitive'):
AttributeError: 'ImportCampaignWindow' object has no attribute 'select_button'
```

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		"""
+		for nods in element.iter():
+			if nods.tag in id_fields and nods.text == old_id:
+				nods.text = new_id
+				# if new_id is none set type to null
+				if new_id is None:
+					nods.attrib['type'] = 'null'
+
+	def _get_keys_values(self, element):
+		keys = []
```
zeroSteiner: Since you were have performance issues with this code, you might get it to run faster if you swapped all three of these lists, keys, rows and values for [collections.deque](https://docs.python.org/2/library/collections.html#collections.deque) instances. Should get faster appends.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+
+	def close_import_campaign(self, _):
+		"""
+		Checks to make sure the import campaign thread is closed before closing.
+		"""
+		if not self.campaign_info:
+			return
+		if not self.thread_import_campaign:
+			return
+		if not self.thread_import_campaign.isAlive():
```
zeroSteiner: Use the newer and pep8 compliant name of `.is_alive` here instead of the camelCase one.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+			return
+
+		self.thread_import_campaign.stop()
+		self.thread_import_campaign.join()
+		self.remove_import_campaign()
+
+	def failed_import_action(self):
+		response = gui_utilities.show_dialog(
+			Gtk.MessageType.QUESTION,
+			'Import Failed, do you want to delete this campaign?',
```
zeroSteiner: `Import failed, ...` "Failed" is not a proper noun or the first word of this sentence.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+
+		GLib.idle_add(self.import_progress.set_fraction, 1.0)
+		GLib.idle_add(self.button_import_campaign.set_sensitive, False)
+		self.campaign_info = None
+		GLib.idle_add(self.entry_campaign_name.set_text, '')
+		GLib.idle_add(self.entry_path.set_text, '')
+		GLib.idle_add(self.select_button.set_sensitive, True)
+		GLib.idle_add(self.spinner.stop)
+		done_string = 'Done importing campaign it took {}'.format(datetime.datetime.now() - start_time)
+		GLib.idle_add(self._update_text_view, done_string)
```
zeroSteiner: This paradigm of using `GLib.idle_add` to call `_update_text_view` seems pretty tedious. If there are no instances where you don't call `_update_text_view` from `GLib.idle_add` it would be cleaner to just put the call into the function. If there are, perhaps adding a kwarg of `idle_add` to the function and having it default to what ever is used most would be beneficial.

docs/source/change_log.rst
```diff
@@ -17,6 +17,7 @@ Version 1.7.0
 * New features for client and server plugins
 * Add comparison of "trained" statistics to the campaign comparison
 * Support for including and importing Jinja templates from relative paths
+* New feature to import Campaigns from XML files.
```
zeroSteiner: No period to be consistent with all of the other lines.

king_phisher/client/windows/campaign_import.py
wolfthefallen: maybe tedious but due to adding text to the end of the buffer, it must be wrapped in the `idle_add` so the true position of the of the end of the buffer at the time of update can be determined.

zeroSteiner: You're referring to `end_iter` right? So this wouldn't work:
```python
    def __update_text_view(self, string_to_add):
        end_iter = self.text_buffer.get_end_iter()
        self.text_buffer.insert(end_iter, string_to_add + "\n")

    def _update_text_view(self, string_to_add, idle=True):
        if idle:
            GLib.idle_add(self.__update_text_view, string_to_add)
        else:
            self.__update_text_view(string_to_add)
        return None
```

king_phisher/client/windows/campaign_import.py
zeroSteiner: This should use the `gui_utilities.show_dialog_yes_no(` function.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		Used to delete the imported campaign on failure or early exit of the import window, if the user
+		selects to have it removed.
+		"""
+		campaign_id = self.campaign_info.find('id').text
+		campaign_name = self.campaign_info.find('name').text
+		campaign_check = self.rpc('db/table/get', 'campaigns', campaign_id)
+		if not campaign_check:
+			return
+		if campaign_name == campaign_check['name']:
+			self.rpc('db/table/delete', 'campaigns', campaign_id)
```
zeroSteiner: Please use the `campaign-delete` signal from the application. See [the docs](http://king-phisher.readthedocs.io/en/latest/client/gobject_signals.html#application-signals) for details. It's marked with the flag `SIGNAL_ACTION` which means it's safe to be invoked.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+			self.logger.info('Deleted Campaign {}'.format(campaign_id))
```
zeroSteiner: Please don't start logging messages with capitals, nothing else does in King Phisher with the exception of acronyms which are inconsistent.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+
+	def signal_import_button(self, _):
+		"""
+		Function called when the user starts the import campaign process, It will create a new thread,
```
zeroSteiner: `...process, It will create a new thread...` is improper grammar, looks like a period is missing.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		to have the campaign run in the background freeing up the GUI for the user to conduct other functions.
```
zeroSteiner: Strictly speaking the campaign is not running in the background, you mean the import thread.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		if not response:
+			return
+		target_file = response['target_path']
+		self.entry_path.set_text(target_file)
+
+		try:
+			campaign_xml = ET.parse(target_file)
+		except ET.ParseError as error:
+			self.logger.error('cannot import campaign: {0} is not a valid XML file'.format(target_file), error)
+			raise KingPhisherInputValidationError('{0} is not a valid xml file'.format(target_file))
```
zeroSteiner: Strings used as format templates should use double quotes. Same with the line above. This makes it easier to visually identify them from editors that highlight them in different colors.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+
+		root = campaign_xml.getroot()
+		if root.tag != 'king_phisher':
+			raise KingPhisherInputValidationError('File not a King Phisher Campaign XML Export File')
```
zeroSteiner: Don't need "File" at the beginning and end of this error message, it's redundant.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		meta_data = root.find('metadata')
+		if meta_data.find('version').text < '1.3':
+			raise KingPhisherInputValidationError('Can only import XML Campaign data version 1.3 or higher')
+		self.campaign_info = root.find('campaign')
+		if not self.campaign_info:
+			raise KingPhisherInputValidationError('XML File does not contain any campaign information')
```
zeroSteiner: "File" does not need to be capitalized here.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		self.entry_campaign_name.set_text(self.campaign_info.find('name').text)
+		self.thread_import_campaign = None
+		if not self._check_campaign_name(self.campaign_info.find('name').text, verbose=True):
+			self.button_import_campaign.set_sensitive(False)
+			return
+		self.button_import_campaign.set_sensitive(True)
+
+	def _check_campaign_name(self, campaign_name, verbose=False):
+		"""
+		Used to check to see the user has provided a campaign name that is not already in use.
```
zeroSteiner: "Used to check to see **if** the user has..."

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		If verbose is used it will display information in detailed output text buffer.
```
zeroSteiner: When referencing a parameter within the functions documentation, I think you're supposed to use ```**``` so sphinx will adjust the formatting.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+
+		:param str campaign_name: campaign name to check
+		:param bool verbose:
```
zeroSteiner: Verbose is not documented here, I know it's in the functions description, maybe that info should be moved here.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+			self.button_import_campaign.set_sensitive(False)
+			return
+		self.button_import_campaign.set_sensitive(True)
+
+	def _check_campaign_name(self, campaign_name, verbose=False):
+		"""
+		Used to check to see the user has provided a campaign name that is not already in use.
+		If verbose is used it will display information in detailed output text buffer.
+
+		:param str campaign_name: campaign name to check
```
zeroSteiner: "The campaign name to check."

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		if next((nodes for nodes in self.db_campaigns if nodes['node']['name'] == campaign_name), None):
+			if verbose:
+				self._update_text_view('Campaign name {} is already in use by another campaign.'.format(campaign_name), idle=True)
+			return False
+
+		if verbose:
+			self._update_text_view('Campaign Name {} is not in use, ready to import'.format(campaign_name), idle=True)
+		return True
+
+	def prepep_xml_data(self):
```
zeroSteiner: This function should be renamed "preprep_xml_data".

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+		"""
+		This function provides the actions required to see if depended IDs are already in the database.
```
zeroSteiner: "... to see if **required** IDs are already..."

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+			if self.thread_import_campaign.stopped():
+				return
+
+			# make table rows easy to manage for updating new ids returned
+			table_rows = []
+			for row in rows:
+				row = dict(zip(keys, row))
+				table_rows.append(row)
+
+			while len(rows) > 0:
```
zeroSteiner: This can simply be ```while rows```.

king_phisher/client/windows/campaign_import.py
```diff
@@ -0,0 +1,439 @@
...
+				if self.thread_import_campaign.stopped():
+					return
+				try:
+					inserted_ids = inserted_ids + self.rpc('/db/table/insert/multi', tables, keys, rows[:batch_size], deconflict_ids=True)
+				except advancedhttpserver.RPCError:
+					response = gui_utilities.glib_idle_add_wait(self.failed_import_action)
+					if response == -8:
```
zeroSteiner: There should either be a constant mapped to `-8` here or it needs a comment regarding it's significance.

zeroSteiner:
Everything checks out now, I made the following changes and will land the PR momentarily.

* Adjusted white space of widgets in the window
* Adjusted strings of "XML Campaign..." to "Campaign XML..."
* Switched to a `Gtk.ApplicationWindow`, **which fixed an odd reference bug**

Thanks @wolfthefallen

PR #201, opened 2017-03-24
Add Additional Default Client Configuration Keys

wolfthefallen:
# Description
~~This PR changes the `merge_config` function to allow new keys to be create and values set during the merging process.~~
Updated base user configuration file to support default keys to facilitate importing common key values from another configuration file, while still maintaining a white list.

Custom profiles/configs not loading required default keys during import causing stack traces to occur later in the program. Primary example at this time is SMTP setting, using ssh connections.

thumbnail_kingphisher-error

```
ERROR    error uid: bb8c1231-8c6d-4cc3-8740-7da020450bf9 an unhandled exception was thrown
  File "/opt/king-phisher/king_phisher/client/tabs/mail.py", line 292, in signal_button_clicked_sender_start
    if not self._sender_precheck_spf():
  File "/opt/king-phisher/king_phisher/client/tabs/mail.py", line 228, in _sender_precheck_spf
    spf_test_ip = mailer.guess_smtp_server_address(self.config['smtp_server'], (self.config['ssh_server'] if self.config['smtp_ssh_enable'] else None))
KeyError: 'ssh_server'
```

# Testing
- [ ] Remove current `.config/king-phisher/config.json` file.
- [ ] On the login window right click and import configuration file, that has keys not set on default load, example key `ssh_server`
- [ ] Login and then then close King Phisher
- [ ] Cat out `~/.config/king-phisher/config.json' to verify import was successful

### DIFF
--- a/data/client/king_phisher/client_config.json
+++ b/data/client/king_phisher/client_config.json
@@ -16,13 +16,17 @@
   "server": "localhost:22",
   "server_remote_port": 80,
   "server_use_ssl": false,
+  "server_username": "",
   "sftp_client": "filezilla --logontype=interactive sftp://{username}@{server}{web_root}",
   "ssh_preferred_key": null,
   "smtp_max_send_rate": 45.0,
   "smtp_server": "localhost:25",
   "smtp_ssh_enable": false,
   "smtp_ssl_enable": false,
+  "smtp_username": "",
   "spf_check_level": 1,
+  "ssh_server": "localhost:22",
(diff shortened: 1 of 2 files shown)


zeroSteiner:
Instead of removing the whitelisting of keys, why not add the key names to the [config file](https://github.com/securestate/king-phisher/blob/master/data/client/king_phisher/client_config.json) that non-existent keys are loaded from when the application starts up?

Issue #206, opened 2017-04-21
Failed to detect the version of linux


drc0nc:
## Issue Description
Failed to detect the version of linux on new install. I'm running the newest ubuntu server 17.04 

## Reproduction Steps
Upon fresh install the script gives the failed to detect version of linux.

## Environment Details

Host OS: Ubuntu 17.04
King Phisher Version: newest pull

## Error Details / Stack Trace

```
If you are opening an issue regarding the client, copy text from the "Error
Details" of the error dialog in the client here. If you are opening an issue
regarding the server copy the relevant lines from the logs here.
```

zeroSteiner:
@wolfthefallen looks like the install script needs to be updated to identify 17.04. This'll be an easy fix.

drc0nc:
I forgot to mention that I'm currently in the dev branch

zeroSteiner:
Fixed in commit c64bbd80b4af54ffb7a696f7ea2bd817350fdf65 which is available in both the `master` and `dev` branches.  Thanks for beta testing!

PR #207, opened 2017-04-21
Updated install.sh

wolfthefallen:
# Install Script update
This PR updates the `tools/install.sh` script to include the latest version of Ubuntu 17.04. This will correct issue #206 

## Testing
- [ ] `sudo tool/install.sh` successfully installs King Phisher on Ubuntu 17.04


zeroSteiner:
@wolfthefallen closing this out since it's a one line change and should actually be targeting the `master` branch due to being a bug. Additionally, master should be bumped to `1.7.2-beta` which I'll take care of at the same time then cherrypick this bug fix back into `dev`.

PR #215, opened 2017-05-15
Added new uninstall.sh that can remove user configuration

hdemeyer:
# Description
I added an option to delete king phisher configuration for a user. The user can be detected automatically or a user can be supplied.

# Usage
`tools/uninstall.sh --delete-config`
`tools/uninstall.sh --delete-config [specific_user]`
- [ ] finds base user running command and promts yes or no to delete profile
- [ ] requests to delete specified user when supplied

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -1,0 +1,201 @@
+#!/bin/bash
+# vim: tabstop=4 softtabstop=4 shiftwidth=4 noexpandtab
+###############################################################################
+# This is the Linux uninstall script for the King Phisher Client and
+# Server on supported distributions.
+#
+# Project Home Page: https://github.com/securestate/king-phisher/
+# Authors:
+#   Erik Daguerre
+#	Hunter DeMeyer
+#
+###############################################################################
+
+E_USAGE=64
+E_SOFTWARE=70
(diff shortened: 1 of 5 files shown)


uninstall.sh
```diff
@@ -0,0 +1,201 @@
+#!/bin/bash
+# vim: tabstop=4 softtabstop=4 shiftwidth=4 noexpandtab
+###############################################################################
+# This is the Linux uninstall script for the King Phisher Client and
+# Server on supported distributions.
+#
+# Project Home Page: https://github.com/securestate/king-phisher/
+# Authors:
+#   Erik Daguerre
+#	Hunter DeMeyer
```
wolfthefallen: remove spaces to line your name up with Erik

zeroSteiner:
This pull request needs to be redone so it is only modifying the `uninstall.sh` file. It looks like the commits from master that are already in dev were included. Contributions should also be done from a topic branch to avoid this in the future.

Save your changes to `uninstall.sh` outside of the project, reset your branch and checkout a topic branch that is synced with `dev`.

zeroSteiner:
Yeah, I don't think your going to be able to rebase your way out of this one.

PR #229, opened 2017-07-23
python3 -V

cclauss:
Are Python versions prior to 3.5 supported?


zeroSteiner:
Yes, 3.3 and up should work. This did change recently, the last release we made is *technically* the last that supports Python 2.7. We'll fix bugs in it but all work moving forward and the next release will be Python 3 exclusively.

PR #234, opened 2017-08-09
Support for Arch Linux

cbelk:
tools/install.sh updated to include support for Arch Linux. Tested install script on a few fresh installs of Arch.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -22,6 +22,7 @@
 #   Kali            | yes    | yes    |
 #   Red Hat         | no     | yes    |
 #   Ubuntu          | yes    | yes    |
+#   Arch            | yes    | yes    |
 #
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


tools/install.sh
```diff
@@ -336,6 +344,18 @@ elif [ "$LINUX_VERSION" == "BackBox" ] || \
 		easy_install -U distribute
 		apt-get install -y postgresql-server-dev-all &> /dev/null
 	fi
+elif [ "$LINUX_VERSION" == "Arch" ]; then
+	pacman --noconfirm -S freetype2 python python-pip pkg-config
```
zeroSteiner: Is this installing Python2 or Python3, because 1.8 (the latest release) is the last one in which we support Python2.

cbelk: That is installing Python3. Python3 is the default Python for Arch (which is why just python is used -- to target Python2 would be python2).

wolfthefallen:
@cbelk Thank you for the pull request

Issue #233, opened 2017-08-02
KingPhisher Clients Unresponsive


0xIn7rud3r:
## Issue Description
KingPhisher Clients gets stuck forever on both linux and windows machines. Kingphisher Server is running on another machine, with client configured to use SSH over tunnel for sending mail

## Reproduction Steps
1. Start King Phisher
2. Do something to trigger the crash

## Environment Details

Host OS:  Kali 2,
KingPhisherServer Version: 1.7.1

## Error Details / Stack Trace
No Error, Just client become unresponsive. till force QUIT

![image](https://user-images.githubusercontent.com/30025517/28872487-1055a1e6-77a7-11e7-80fa-b3729555372b.png)

zeroSteiner:
Can you update to the latest version 1.8 and see if you can reproduce this issue. If you can reproduce it, please provide us with the [verbose output](https://github.com/securestate/king-phisher/wiki/Frequently-Asked-Questions#how-do-i-generate-verbose-logging-output).

0xIn7rud3r:
I tried with Updated Server i.e 1.8 and windows client and client becomes unresponsive as soon as send is clicked.

![image](https://user-images.githubusercontent.com/30025517/28874243-a474feec-77ae-11e7-82a2-fa05ec07e376.png)

zeroSteiner:
Ok please provide us with the logs from when you start it until it freezes. You can put them on pastebin or another service if you'd like. I'm expecting it may be a few hundred lines but it's important that we get that information.

Additionally, is the URL up? I'd expect the client may become unresponsive for a few seconds as it waits for the request to complete but no more than maybe 30 seconds or so.

0xIn7rud3r:
I reconfigured all from scratch, everything seems to be working fine now. i'll update with logs if i am able to reproduce it again.

0xIn7rud3r:
So Client becomes unresponsive as soon as start is clicked. 
Running KP server on another host then client.
Client is configured to use SMTP over SSH channel.

Few points i noticed.

No pop for ssh login after clicking Start in send tab. client becomes unresponsive.
Also getting into not responding state when clicking check SPF.
![image](https://user-images.githubusercontent.com/30025517/28916421-ca85c57e-785f-11e7-96aa-9dac46a1dd59.png)


Note: URL is up and running fine. SPF records are also file as i was able to send some mails earlier. now its getting stuck for same client config.

[...]

PR #235, opened 2017-08-10
DNS timeout issues

wolfthefallen:
toThis PR addresses issue #233 where the client become unresponive during SPF checks. With this PR King Phisher will now time out by default at 10 seconds if it has not recieved a response to the DNS query. This option can be changed in the client configuration file with the option `spf_check_timeout`. Furthermore this PR update the SPF query to try all DNS servers. If they fail, it will move to the next, and then remove all DNS servers from its list to one that successfully made response to speed up the process of SPF checking.

To validate:
- [ ] Run King Phisher send an email should work normally
- [ ] Change Primary DNS server to a fake server or ip address that is not a DNS server
- [ ] Run King Phisher will 'hang' for a few seconds then a dialog box will ask say that a timeout occurred and ask the user if you want to continue.
- [ ] Changing the `spf_check_timeout` changes the timeout period when querying DNS server.

### DIFF
--- a/king_phisher/spf.py
+++ b/king_phisher/spf.py
@@ -59,6 +59,11 @@ The maximum number of DNS queries allowed to either return with rcode 0 and no
 answers or rcode 3 (Name Error) as defined within section 4.6.4 of :rfc:`7208`.
 """
 
+DEFAULT_DNS_TIMEOUT = 10
+"""
+The default number of seconds to wait for a query response from the DNS server.
+"""
+
 QUALIFIERS = {
 	'+': SPFResult.PASS,
 	'-': SPFResult.FAIL,
@@ -170,15 +175,15 @@ class SPFTempError(SPFError):
 	"""
 	pass
 
-class SPFTimeOutError(SPFError):
+class SPFTimeOutError(SPFTempError):
 		"""
-		Exception indicating that a timeout occurred querying the DNS server.
-		This is normally caused when the client communicate with the DNS server.
+		Exception indicating that a timeout occurred while querying the DNS server.
+		This is normally caused when the client can't communicate with the DNS server.
 		"""
 		pass
 
 @smoke_zephyr.utilities.Cache('3m')
-def check_host(ip, domain, timeout=10, sender=None):
+def check_host(ip, domain, timeout=DEFAULT_DNS_TIMEOUT, sender=None):
 	"""
 	Analyze the Sender Policy Framework of a domain by creating a
 	:py:class:`.SenderPolicyFramework` instance and returning the result of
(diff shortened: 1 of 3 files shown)


king_phisher/spf.py
```diff
@@ -170,8 +170,15 @@ class SPFTempError(SPFError):
 	"""
 	pass
 
+class SPFTimeOutError(SPFError):
```
zeroSteiner: This should inherit from `SPFTempError`.

king_phisher/spf.py
```diff
@@ -170,8 +170,15 @@ class SPFTempError(SPFError):
...
+		"""
+		Exception indicating that a timeout occurred querying the DNS server.
```
zeroSteiner: `...occurred *while* querying the DNS server.`

king_phisher/spf.py
```diff
@@ -170,8 +170,15 @@ class SPFTempError(SPFError):
...
+		This is normally caused when the client communicate with the DNS server.
```
zeroSteiner: `...when the client *can't* communicate with the ...`

king_phisher/spf.py
```diff
@@ -208,11 +216,12 @@ class SenderPolicyFramework(object):
 	if an IP address is authorized to send messages on it's behalf. The exp
 	modifier defined in section 6.2 of the RFC is not supported.
 	"""
-	def __init__(self, ip, domain, sender=None):
+	def __init__(self, ip, domain, timeout, sender=None):
```
zeroSteiner: Should probably be a keyword arg here as well. To keep them both synced with the same value, you should add a constant to the top of the file and document it with the others. Something like `DEFAULT_DNS_TIMEOUT = 10` and them `timeout=DEFAULT_DNS_TIMEOUT`.

king_phisher/spf.py
```diff
@@ -330,31 +343,47 @@ def _check_host(self, ip, domain, sender, top_level=True):
...
+			answers = []
+			if len(response.answer) == 0 or rcode == dns.rcode.NXDOMAIN:
+				self.logger.debug("resolving {0:<3} record for {1} using nameserver {2} resulted in a void lookup".format(qtype, qname, nameserver))
+				self.query_limit_void -= 1
+				if self.query_limit_void < 0:
+					raise SPFPermError('DNS query void lookup limit reached')
+			for answer in response.answer:
+				answers.extend(answer.items)
+
+			if self.nameservers.index(nameserver) != 0:
```
zeroSteiner: The `!= 0` is redundant. Any integer other than 0 evaluates to True.

zeroSteiner:
Landed with some modifications. Thanks @wolfthefallen !

PR #269, opened 2018-03-13
Update Requirements.txt

y4utj4:
matplotlib 2.2.0 was throwing a nasty stack trace, using 2.1 works just find.


zeroSteiner:
Change makes sense, I'm going to merge this in now.

PR #279, opened 2018-03-28
Install script

wolfthefallen:
This PR includes 2 updates the `tools/install.sh` script.

First the user can now manually select the type of *nix system they are utilizing if the auto detect fails.

Second the user can now have the database uri for PostgreSQL from their original `server_config.yml` file to be applied to the new `server_config.yml` if running the install script again for updates.

#  Manually Select *nix testing
- [x] If the *nix os you are currently testing with between lines 147-183 and change the regex so your os is no longer supported.
- [x] Run `sudo tools/install.sh`
- [ ] It promts you to select 1-7 for your selected operating system
   - [x] If an old server_config.yml is found in the directory you are installing from see Keep Database uri
- [x] Script attempts to install King Phisher

# Keep Database URI testing
- [x] Put the new install script in directory `./tools' directory of current King Phisher install
- [x] run `sudo tools/install.sh`
- [x] Script warns you that it has found an old `server_config.yml` file
- [x] Script asks if you want to use old database uri
- [x] If yes it puts it in new `server_config.yml'
- [x] if no it does not put the database uri in the new 'server_config.yml'


zeroSteiner:
These updates work great. Fantastic stuff right here, I particularly like the maintaining of the database connection string.

I'll get this landed in a minute, I made a few grammar and spelling changes. I also updated almost all of the echo statements to be prefixed with `INFO:`, `WARNING:`, or `ERROR:`. I think these changes are changelog worthy so I'll get a note in there as well.

PR #285, opened 2018-03-30
Catalog reload

wolfthefallen:
This PR updates the plugin to allow for the user to force reload the selected catalog.
If a repository is selected it will reload the top level catalog.
Updated install/uninstall methods to display information on status bar, so it is more verbose to end user.

To test
- [x] Open Plugin Manager
- [x] select a catalog row, right click and select `reload`
  - [x] Selected catalog is removed and reloaded.
- [x] select a repository row, right click and select `reload`
  - [x] Repository's top level catalog is removed and reloaded.
- [x] select an installed plugin, right click and select `reload`
  - [x] selected plugin reloads

### DIFF
--- a/king_phisher/client/windows/plugin_manager.py
+++ b/king_phisher/client/windows/plugin_manager.py
@@ -185,6 +185,7 @@ class PluginManagerWindow(gui_utilities.GladeGObject):
 		self._load_plugins()
 
 	def _load_catalog_from_url(self, catalog_url):
+		catalog = None
 		try:
 			catalog = Catalog.from_url(catalog_url)
 		except requests.exceptions.ConnectionError:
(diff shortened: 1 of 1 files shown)


king_phisher/client/windows/plugin_manager.py
```diff
@@ -178,18 +178,23 @@ def _load_catalogs(self, refresh=False):
...
+	def _load_catalog_from_url(self, catalog_url):
+		try:
+			catalog = Catalog.from_url(catalog_url)
+		except requests.exceptions.ConnectionError:
+			self.logger.warning("connection error trying to download catalog url: {}".format(catalog_url))
+			self.idle_show_dialog_error('Catalog Loading Error', "Failed to download catalog, check your internet connection.")
+		except Exception:
+			self.logger.warning("failed to add catalog by url", exc_info=True)
+			self.idle_show_dialog_error('Catalog Loading Error', "Failed to add catalog")
+		return catalog
```
zeroSteiner: This should be in an else block or the except blocks should end in `return None` otherwise `return catalog` will error out because `catalog` is undefined.

king_phisher/client/windows/plugin_manager.py
```diff
@@ -287,6 +284,17 @@ def _load_plugins(self):
 		gui_utilities.glib_idle_add_once(self._treeview_unselect)
 		self._update_status_bar('Loading completed', idle=True)
 
+	def _add_catalog_to_tree(self, catalog_id, store):
+		model = (catalog_id, None, True, catalog_id, None, None, False, False, False, _ROW_TYPE_CATALOG)
```
zeroSteiner: This should be using the named tuple.

king_phisher/client/windows/plugin_manager.py
```diff
@@ -287,6 +284,17 @@ def _load_plugins(self):
...
+		catalog_row = gui_utilities.glib_idle_add_wait(self._store_append, store, None, model)
+		for repo in self.catalog_plugins.get_repositories(catalog_id):
+			model = (repo.id, None, True, repo.title, None, None, False, False, False, _ROW_TYPE_REPOSITORY)
```
zeroSteiner: Same here, named tuple please.

king_phisher/client/windows/plugin_manager.py
```diff
@@ -413,14 +427,23 @@ def signal_popup_menu_activate_reload(self, _):
 		for tree_iter in gui_utilities.gtk_treeview_selection_iterate(treeview):
 			model_row = self._model[tree_iter]
 			# only reloading installed plugins is currently supported
-			if model_row.parent is None:
-				continue
 			named_row = self._named_model(*model_row)
 			if named_row.type != _ROW_TYPE_PLUGIN:
```
zeroSteiner: Can you make this block a bit more explicit. We have the type information already in the `named_row`, it seems like we should be using to check if it's a catalog and proceeding or if it's a repository find it's parent catalog and proceeding.

zeroSteiner:
Landed, thanks @wolfthefallen !

PR #297, opened 2018-05-26
Updated db initailization failed message to be more descriptive

jacobj10:
I'm not really sure you're guys convention for error handling so I figured I'd just throw this out there, spent about 10 minutes figuring out that my DB was not configured correctly. "initialization failed" is a very generic error message to return to the top-level, but "database initialization failed" would at least give me a better idea of where to start looking. Plus, all other instance of `KingPhisherDatabaseError` appeared to give some indication it was DB related in the message.


zeroSteiner:
This looks great thanks @jacobj10! I'll get it landed shortly.

PR #303, opened 2018-06-07
Add new issue templates

zeroSteiner:
This uses the new Github issue templates.

Steps to test:
- [ ] Read the bug report template and ensure it makes sense (spoiler alert, it's the same one we've used previously)
- [ ] Read the new feature request template and ensure it makes sense, propose changes and improvements as necessary

We can add more templates in the future as necessary. I tried to avoid using the word King Phisher in the new one so we can use it for other projects as well.


zeroSteiner:
@wolfthefallen feedback implemented in commit ad96624.

wolfthefallen:
@zeroSteiner thank you for updating the templates!

PR #305, opened 2018-06-18
added optional entry box to show_dialog function

jacobj10:
Still working on catalogs, but figured I'd get some feedback on this small change in the meantime.

Instead of creating a new function for displaying text, I simply added some parameters to facilitate this.

### DIFF
--- a/king_phisher/client/gui_utilities.py
+++ b/king_phisher/client/gui_utilities.py
@@ -466,7 +466,7 @@ def gtk_widget_destroy_children(widget):
 	for child in widget.get_children():
 		child.destroy()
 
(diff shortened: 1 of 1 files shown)


zeroSteiner:
Instead of requiring a list could you support a call back function? A list's append method could be used but it would grant more flexibility in the long term. We could also look for cleaner options than relying on a mutable container.

jacobj10:
my other thought would be to add a generic callback call back function to show_dialog, creating a more flexible way to modify the properties of a dialog before showing?

zeroSteiner:
I like that idea. It'd be better if there are more custom use cases we haven't thought of but would add over head for text entries specifically.

PR #317, opened 2018-08-27
New/time selector

True-Demon:
This creates two new classes under the king_phisher/client/widget/managers.py for the purpose of creating a re-usable, flexible, and general-purpose time selector widget that returns a basic `datetime.time(hour, minute)` object. 

A test example has been implemented in the campaign manager and verified as functional by comparing the default value in the campaign_assistant.expiration value (set to `datetime.time(0, 0)`) and the value that is replaced after the campaign configuration has been applied with a new time value.

Complete with fancy pop up & down transitions 🎆 

Convention for implementing the time selector inside a new object should be as follows ...

```
class SomeAssistant:
   dependencies=(
      children = (
         'your_children',
         'your_grand_children',
         'togglebutton_something_time'
      ),
      top_level=(
         'ClockHourAdjustment',
         'ClockMinuteAdjustment'
      )
   )

   def __init__(self, *args, **kwargs):
      ... your definitions ...
      self.time_selector = managers.TimeSelectorButtonManager(self.gobjects['togglebutton_something_time'], self.application)

   def apply_stuff(self):
      some_date_time = (self.some_date, self.time_selector.time)
```

Any references to the attributes inside of the top class object should be done as such
```
time_selector.time       # The time(hr, min) values, returned as a datetime.time() object
time_selector.button    # The button which contains the popover
time_selector.popover.popover      # The actual popover gobject is nested within the TimeSelector.popover object, so watch out for that.
```

### DIFF
--- a/king_phisher/client/assistants/campaign.py
+++ b/king_phisher/client/assistants/campaign.py
@@ -119,7 +119,6 @@ class CampaignAssistant(gui_utilities.GladeGObject):
 			self.gobjects['label_intro_body'].set_text('This assistant will walk you through creating and configuring a new King Phisher campaign.')
 			self.gobjects['label_intro_title'].set_text('New Campaign')
 
-
 	@property
 	def campaign_name(self):
 		"""
@@ -187,8 +186,6 @@ class CampaignAssistant(gui_utilities.GladeGObject):
 			expiration = utilities.datetime_utc_to_local(campaign['expiration'])
 			self.gobjects['checkbutton_expire_campaign'].set_active(True)
 			gui_utilities.gtk_calendar_set_pydate(self.gobjects['calendar_campaign_expiration'], expiration.date())
-
-			
 
 	def _get_tag_from_combobox(self, combobox, db_table):
 		model = combobox.get_model()
--- a/king_phisher/client/assistants/campaign.py
+++ b/king_phisher/client/assistants/campaign.py
@@ -35,6 +35,7 @@ import datetime
 from king_phisher import utilities
 from king_phisher.client import gui_utilities
 from king_phisher.client.widget import resources
+from king_phisher.client.widget import managers
 
 import advancedhttpserver
 
@@ -73,8 +74,7 @@ class CampaignAssistant(gui_utilities.GladeGObject):
 			'radiobutton_company_existing',
 			'radiobutton_company_new',
 			'radiobutton_company_none',
-			'spinbutton_campaign_expiration_hour',
-			'spinbutton_campaign_expiration_minute'
+			'togglebutton_expiration_time'
 		),
 		top_level=(
 			'ClockHourAdjustment',
@@ -99,6 +99,7 @@ class CampaignAssistant(gui_utilities.GladeGObject):
 			if page_title:
 				self._page_titles[page_title] = page_n
 
+		self.time_selector = managers.TimeSelectorButtonManager(self.gobjects['togglebutton_expiration_time'], self.application)
 		self._set_comboboxes()
 		self._set_defaults()
 
@@ -117,6 +118,7 @@ class CampaignAssistant(gui_utilities.GladeGObject):
 			self.gobjects['label_confirm_body'].set_text(confirm_preamble + ', then hit "Apply" to create the new King Phisher campaign.')
 			self.gobjects['label_intro_body'].set_text('This assistant will walk you through creating and configuring a new King Phisher campaign.')
 			self.gobjects['label_intro_title'].set_text('New Campaign')
+
 
 	@property
 	def campaign_name(self):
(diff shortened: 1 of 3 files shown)


king_phisher/client/widget/managers.py
```diff
@@ -373,3 +374,61 @@ def signal_activate_popup_menu_copy(self, menuitem, column_ids):
 
 	def signal_activate_popup_menu_delete(self, menuitem):
 		self._call_cb_delete()
+
+class TimeSelector(gui_utilities.GladeGObject):
+	dependencies = gui_utilities.GladeDependencies(
```
wolfthefallen: add in description for the class, as this is designed to be used in other parts of the application if we ever need to select time again

king_phisher/client/widget/managers.py
```diff
@@ -373,3 +374,61 @@ def signal_activate_popup_menu_copy(self, menuitem, column_ids):
...
+		:param application: The application instance which this object belongs to.
+		:param value: The present datetime value (initialized to 00:00)
+		:type datetime.time: The value, (initialized to NoneType) will always return datetime.time(hr, min)
+		"""
+		self.popover = TimeSelector(application)
+		self.button = button
+		self.application = application
+		self._hour_spin = self.popover.gobjects['spinbutton_hour']
+		self._minute_spin = self.popover.gobjects['spinbutton_minute']
+		self.time = value or datetime.time(0, 0)
```
wolfthefallen: to be on the safe side, I would recommed checking that the value is a datetime.time object prior to setting it to self.time.

king_phisher/client/widget/managers.py
```diff
@@ -373,3 +374,61 @@ def signal_activate_popup_menu_copy(self, menuitem, column_ids):
...
+		self.button.connect('toggled', self.signal_button_toggled)
+		self.popover.popover.set_relative_to(self.button)
+		self.button.set_label(f"{0:02}:{0:02}")
+
+	def __repr__(self):
+		return "<{0} time='{1:%H:%M}' >".format(self.__class__.__name__, self.time)
```
wolfthefallen: is there a way to get the name with out calling the private objects on this class?

king_phisher/client/widget/managers.py
```diff
@@ -373,3 +374,61 @@ def signal_activate_popup_menu_copy(self, menuitem, column_ids):
...
+
+	def signal_button_toggled(self, _):
+		if not self.button.get_active():
```
wolfthefallen: flip the logic save some space

king_phisher/client/widget/managers.py
```diff
@@ -373,3 +374,61 @@ def signal_activate_popup_menu_copy(self, menuitem, column_ids):
...
+		top_level=(
+			'ClockHourAdjustment',
+			'ClockMinuteAdjustment'
+		)
+	)
+	top_gobject = 'popover'
+
+class TimeSelectorButtonManager(object):
+	def __init__(self, button, application, value=None):
+		"""
```
wolfthefallen: I would add a description here of what this class is used for. The parameter descriptions are good.

king_phisher/client/assistants/campaign.py
```diff
@@ -249,6 +248,9 @@ def _get_graphql_company(self, company_name):
 		}""", {'name': company_name})
 		return results['db']['company']
 
+	def signal_toggle_time(self, _):
+		return self.time_selector.signal_button_toggled(_)
```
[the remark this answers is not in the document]
True-Demon: I'll wait for @zeroSteiner's input because I think I remember him telling me to explicitly call it this way, the same way the `signal_assistant_apply(self, _)` is called.

king_phisher/client/widget/managers.py
True-Demon: Don't know. This was @zeroSteiner's implementation again.

king_phisher/client/widget/managers.py
True-Demon: This is accomplished by the `value or datetime.time(0, 0)` because value will either be a NoneType or a datetime when it is read. If it evaluates a NoneType, it automatically reassigns it to `datetime.time(0, 0)` so the check is not necessary.
Thanks to @zeroSteiner for that little bit of py-fu.

king_phisher/client/widget/managers.py
True-Demon: I'll add the description, but in this case, the class is reserved to only be used with the TimeSelectorButtonManager. This is the code that instantiates the TimeSelector as a GTK.Popover() and gets fed directly into `TimeSelectorButtonManger(button=TimeSelector, application=self.application))`

king_phisher/client/widget/managers.py
True-Demon: I tried this initially, and it caused a weird loop where it would always be stuck open. I don't know why that occurred, but this was how I resolved it.

zeroSteiner:
Landed with a few changes in commit 3adcef5f83cda93f04517ea42297630c88f17927.

zeroSteiner:
Thanks @True-Demon! This looks so much better.

PR #325, opened 2018-10-08
pipenv enviroment

wolfthefallen:
This PR creates a python virtual environment to run King Phisher in.

This is done by utilizing [pipenv](https://pipenv.readthedocs.io/en/latest/). This python library makes it easy for configuration of both a production, development environments. As will as supporting different versions of python all contained within their own environment. Making this change allows easier development on the King Phisher project while making a more ease of use for end users, as it does not effect any python packages that may have been installed with their package manager, or breaking other python programs by changing system package versions.

`tools/install.sh` will now install all required system packages and then call `./KingPhisher` to set up the pipenv environment.

In addition to this, when starting King Phisher's Client or Server if the virtual environment is not found it will set it up prior to launching the application.

As there is no longer a `requirements.txt` file, upgrading is now handled by `./KingPhisher` or `./KingPhisherServer` with the switch `--env_update`. Which will call `pipenv` to sync, recreate the `Pipefile.lock` and reinstall the virtual environment.

Please Note at the time of this PR their is an issue with pipenv 2018.7.1 and pip 18.1. The issue has been resolved with 2018.7.1.dev0, but is not released with pypi. The installation script and pipenv wrapper for the client and server will install pipenv directly from github as a work around. Once the next version of pipenv is released to pypi this should be revisited.

To support having a wrapper for King Phisher's applications, a new utilities file has been created called `clean_utilities.py` this file should only contain functions and or classes that does only utilizes native python libraries. As this is imported for key functionality of the wrapper prior to the virtual environment being set up. 

testing:
- [x] `tools/install.sh` installs King Phisher on supported operating systems (I have tested it on Ubuntu 18, Fedora, and Kali rolling)
- [x] './KingPhisher` and `/KingPhisherServer` will check for the virtual enviroment and if not call `pipenv` to install it.
- [x] './KingPhisher` and `/KingPhisherServer` `--env_update` and `--env_install` update/install the pipenv environment and cleanly exits.


king_phisher/server/__main__.py
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/python3 -B
```
zeroSteiner: Same here, these first 4 lines should be updated.

king_phisher/clean_utilities.py
```diff
@@ -0,0 +1,84 @@
...
+	for path in os.environ['PATH'].split(os.pathsep):
+		path = path.strip('"')
+		exe_file = os.path.join(path, program)
+		if is_exe(exe_file):
+			return exe_file
+	if is_exe(program):
+		return os.path.abspath(program)
+	return None
+
+def argp_add_defualt_args(parser, default_root=''):
```
zeroSteiner: You misspelled "default" in the function name, but oddly enough it's spelled correctly in the keyword argument.

tools/development/pipenv_wrapper
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/python3 -B
```
zeroSteiner: I don't see where this file is referenced at all or why it is necessary. It seems redundant.

KingPhisher
```diff
@@ -34,74 +34,59 @@ import argparse
...
+
+	target_directory = os.path.abspath(os.path.dirname(__file__))
+	logger.debug("target diretory: {}".format(target_directory))
+	os.environ['PIPENV_VENV_IN_PROJECT'] = os.environ.get('PIPENV_VENV_IN_PROJECT', 'True')
+	os.environ['PIPENV_PIPFILE'] = os.environ.get('PIPENV_PIPFILE', str(os.path.join(target_directory, 'Pipfile')))
+	os.chdir(target_directory)
+	logger.info('checking for pipenv for environment')
+	if not clean_utilities.which('pipenv'):
+		logger.info('installing pipenv')
+		subprocess.call(['python3', '-m', 'pip', 'install', 'git+https://github.com/pypa/pipenv.git'])
```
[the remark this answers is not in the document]
zeroSteiner: Also is there a reason, pipenv can't just be installed from Pip using `pip install pipenv`?

king_phisher/clean_utilities.py
```diff
@@ -0,0 +1,84 @@
...
+	kps_group = parser.add_argument_group('King Phisher Server')
+	kps_group.add_argument('-f', '--foreground', dest='foreground', action='store_true', default=False, help='run in the foreground (do not fork)')
+	kps_group.add_argument('--update-geoip-db', dest='update_geoip_db', action='store_true', default=False, help='update the geoip database and exit')
+	kps_group.add_argument('--verify-config', dest='verify_config', action='store_true', default=False, help='verify the configuration and exit')
+	kps_group.add_argument('config_file', action='store', help='configuration file to use')
+	return parser
+
+def argp_add_wrapper(parser):
+	parser.add_argument('--env_update', dest='pipenv_update', default=False, action='store_true', help='update pipenv requirments and exit')
+	parser.add_argument('--env_install', dest='pipenv_install', default=False, action='store_true', help='install pipenv enviroment and exit')
```
zeroSteiner: The word environment is misspelled here.

KingPhisher
```diff
@@ -33,75 +33,75 @@
...
+	target_directory = os.path.abspath(os.path.dirname(__file__))
+	logger.debug("target diretory: {}".format(target_directory))
+	os.environ['PIPENV_VENV_IN_PROJECT'] = os.environ.get('PIPENV_VENV_IN_PROJECT', 'True')
+	os.environ['PIPENV_PIPFILE'] = os.environ.get('PIPENV_PIPFILE', os.path.join(target_directory, 'Pipfile'))
+	logger.info('checking for pipenv for environment')
+	if not startup.which('pipenv'):
+		logger.info('installing pipenv')
+		process_output, return_code = startup.run_process([sys.executable, '-m', 'pip', 'install', 'pipenv'], cwd=target_directory)
+		if return_code:
+			logger.warning("the following issue occurred during installation of pipenv: {}".format(process_output))
```
zeroSteiner: Since we're not continuing here, this should be an error and you should be returning non-zero. I think returning `return_code` would actually be a good idea so the error status "bubbles up".

KingPhisher
```diff
@@ -33,75 +33,75 @@
...
 def main():
 	parser = argparse.ArgumentParser(description='King Phisher Client GUI', conflict_handler='resolve')
-	utilities.argp_add_args(parser, default_root='KingPhisher')
-	parser.add_argument('-c', '--config', dest='config_file', required=False, help='specify a configuration file to use')
-	parser.add_argument('--no-plugins', dest='use_plugins', default=True, action='store_false', help='disable all plugins')
-	parser.add_argument('--no-style', dest='use_style', default=True, action='store_false', help='disable interface styling')
-	arguments = parser.parse_args()
+	startup.argp_add_wrapper(parser)
+	startup.argp_add_client(parser)
+	startup.argb_add_default_args(parser)
```
zeroSteiner: It looks like `argb_add_default_args` should be `argp_add_default_args`. Not sure why that one is argB.

KingPhisher
```diff
@@ -33,75 +33,75 @@
...
-		return 0
-	logger.debug("client loaded in {0:.2f} seconds".format(time.time() - start_time))
+	if not os.path.isdir(os.path.join(target_directory, '.venv')):
+		logger.info("installing King Phisher's pipenv environment")
+		process_output, return_code = startup.run_process([pipenv_path, '--site-packages', 'install'], cwd=target_directory)
+		if return_code:
+			logger.warning("the following error occurred during pipenv environment setup: {}".format(process_output))
+			return 0
+	logger.debug('pipenv Pipfile: {}'.format(os.environ['PIPENV_PIPFILE']))
+	passing_argv = [' ', 'run', os.path.basename(__file__)] + sys_argv
```
zeroSteiner: Add a comment here that the `' '` is necessary.

king_phisher/client/__main__.py
```diff
@@ -0,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+#  client/__main__.py
```
zeroSteiner: Should actually be `king_phisher/client/__main__.py`.

king_phisher/server/__main__.py
```diff
@@ -0,0 +1,243 @@
+# -*- coding: utf-8 -*-
+#
+#  server/__main__.py
```
zeroSteiner: Should actually be `king_phisher/server/__main__.py`.

king_phisher/startup.py
```diff
@@ -0,0 +1,90 @@
...
+	kps_group = parser.add_argument_group('King Phisher Server')
+	kps_group.add_argument('-f', '--foreground', dest='foreground', action='store_true', default=False, help='run in the foreground (do not fork)')
+	kps_group.add_argument('--update-geoip-db', dest='update_geoip_db', action='store_true', default=False, help='update the geoip database and exit')
+	kps_group.add_argument('--verify-config', dest='verify_config', action='store_true', default=False, help='verify the configuration and exit')
+	kps_group.add_argument('config_file', action='store', help='configuration file to use')
+	return parser
+
+def argp_add_wrapper(parser):
+	kpw_group = parser.add_argument_group('King Phisher pipenv wrapper')
+	kpw_group.add_argument('--env-update', dest='pipenv_update', default=False, action='store_true', help='update pipenv requirments and exit')
```
zeroSteiner: The word "requirments" is misspelled, it should be "requirements".

zeroSteiner:
Landed with some minor changes in 2605856.

PR #332, opened 2018-10-26
Windows build fixs

wolfthefallen:
This PR fixes a couple of issues and makes improvements for the windows build.

* an incompatibility use issue with close_fds in the version file. It will now be false if found to be on windows.
* will now default to Gtk style CSSv1 file if on windows, no more manually changing files
* for `build_msi.bat` added in error level checking for build commands, allowing it to cleanly exit the script and report the error code.
* `cx_freeze.py` will now use `re.sub` to get version from the version file, as `cx_freeze` does not allow letters to be in versions.
* `cx_freeze.py` is now set to be used from the development tools folder, and removed goes from manually being bundled.

### DIFF
--- a/tools/development/cx_freeze.py
+++ b/tools/development/cx_freeze.py
@@ -31,10 +31,11 @@
 #
 
 import os
+import re
 import site
 import sys
 
-sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
+sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..')))
 
 from king_phisher import version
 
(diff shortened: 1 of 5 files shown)


tools/development/cx_freeze.py
```diff
@@ -203,7 +196,7 @@
 setup(
 	name='KingPhisher',
 	author='SecureState',
-	version=version.distutils_version,
+	version=re.sub("[^0-9.]", "", version.distutils_version),
```
zeroSteiner: This could be in accurate. By simply removing the letters, the version number isn't always the same value. Take 1.1.0b7 (beta 7) for example would be 1.1.07 which if you convert those to digits for a comparison would be 1.1.7.

Please use version.version_info instead.

zeroSteiner:
Just the one comment, the rest looks good.

wolfthefallen:
@zeroSteiner Its ready for ya ;)

zeroSteiner:
Landed, thanks @wolfthefallen

PR #333, opened 2018-11-06
Fedora29 and Ubuntu18 headless install fixes

wolfthefallen:
This PR fixes the issues when installing issues on Ubuntu 18 Headless servers and Fedora 29.

For Ubuntu 18.04 Headless, python3-pip is not available via `apt-get`. The change here will first try to install `pip3` via `apt-get` if it fails then it will `wget` pypi `get-pip.py` install file and install pip via that method.

For Fedora 29 the `dnf` install packages where adjust for the client install. `vte3` is not available but not required so it was moved to its own line so it does not prevent the installation of other required packages. The next issue was they `pyproj` was not installing correctly due a [known issue](https://github.com/jswhit/pyproj/issues/136#issuecomment-410546181). To fix this issue, the latest `Cython` was added to the Pipefile, and `pyproj` is now install from github as the fix has not been pushed down to pypi.

Testing:
- [ ] `tools/install.sh` now installs on:
  - [ ] with `--skip-client` Ubuntu 18.04 server (headless)
  - [ ] Fedora 29
- [ ] unit tests run successful on both systems
- [ ] Server starts and runs successfully on both operating systems
- [ ] Client starts and runs on Fedora 29

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -159,11 +159,10 @@ function sync_dependencies {
 		apt-get install -y libfreetype6-dev python3-dev pkg-config
 		if ! python3 -m pip --version; then
 			if apt-get install python3-pip; then
-				echo "Installed python3-pip via apt-get"
+				echo "INFO: Installed python3-pip via apt-get"
 			else
-				curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
-				python3 get-pip.py
-				echo "Installed pip via get-pip.py"
+				curl https://bootstrap.pypa.io/get-pip.py | python3
+				echo "INFO: Installed pip via get-pip.py"
 			fi
 		fi
 		if [ -z "$KING_PHISHER_SKIP_CLIENT" ]; then
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -145,7 +145,8 @@ function sync_dependencies {
 			libpng-devel postgresql-devel python3-devel python3-pip \
 			libffi-devel openssl-devel
 		if [ -z "$KING_PHISHER_SKIP_CLIENT" ]; then
-			dnf install -y geos geos-devel gtksourceview3 vte3 gobject-introspection-devel
+			dnf install -y geos geos-devel gtksourceview3 gobject-introspection-devel
+			dnf install -y vte3
 		fi
 		# Fedora 23 is missing an rpm lib required, check to see if it has been installed.
 		if [ ! -d "$/usr/lib/rpm/redhat/redhat-hardened-cc1" ]; then
@@ -155,7 +156,16 @@ function sync_dependencies {
 		 [ "$LINUX_VERSION" == "Debian"  ] || \
 		 [ "$LINUX_VERSION" == "Kali"    ] || \
 		 [ "$LINUX_VERSION" == "Ubuntu"  ]; then
-		apt-get install -y libfreetype6-dev python3-dev python3-pip pkg-config
+		apt-get install -y libfreetype6-dev python3-dev pkg-config
+		if ! python3 -m pip --version; then
+			if apt-get install python3-pip; then
+				echo "Installed python3-pip via apt-get"
+			else
+				curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+				python3 get-pip.py
+				echo "Installed pip via get-pip.py"
+			fi
+		fi
 		if [ -z "$KING_PHISHER_SKIP_CLIENT" ]; then
 			if ! apt-get install -y gir1.2-gtk-3.0 gir1.2-gtksource-3.0 \
 				gir1.2-webkit-3.0 python3-cairo libgeos++-dev libgirepository1.0-dev \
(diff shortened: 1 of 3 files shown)


tools/install.sh
```diff
@@ -155,7 +156,16 @@ function sync_dependencies {
 		 [ "$LINUX_VERSION" == "Debian"  ] || \
 		 [ "$LINUX_VERSION" == "Kali"    ] || \
 		 [ "$LINUX_VERSION" == "Ubuntu"  ]; then
-		apt-get install -y libfreetype6-dev python3-dev python3-pip pkg-config
+		apt-get install -y libfreetype6-dev python3-dev pkg-config
+		if ! python3 -m pip --version; then
+			if apt-get install python3-pip; then
+				echo "Installed python3-pip via apt-get"
```
zeroSteiner: This message needs to be prefixed like all of the other ones are.

tools/install.sh
```diff
@@ -155,7 +156,16 @@ function sync_dependencies {
...
+			else
+				curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
```
zeroSteiner: Is it possible to validate the data coming from PyPa.io so we're not just executing some random script as root? Also can we either delete the file after we're done or pipe it straight into Python?

tools/install.sh
```diff
@@ -155,7 +156,16 @@ function sync_dependencies {
...
+				python3 get-pip.py
+				echo "Installed pip via get-pip.py"
```
zeroSteiner: This message also needs to be prefixed like all the other ones are.

tools/install.sh
```diff
@@ -145,7 +145,8 @@ function sync_dependencies {
 			libpng-devel postgresql-devel python3-devel python3-pip \
 			libffi-devel openssl-devel
 		if [ -z "$KING_PHISHER_SKIP_CLIENT" ]; then
-			dnf install -y geos geos-devel gtksourceview3 vte3 gobject-introspection-devel
```
zeroSteiner: Did you test that removing `vte3` doesn't break the ability to install on older versions of Fedora?

wolfthefallen: yup this is why `vte3` is on its own line. This allows all the other required packages for 29 to be installed, and then an attempt to install `vte3` in case it is an older version of Fedora.

tools/install.sh
```diff
@@ -155,7 +156,15 @@ function sync_dependencies {
 		 [ "$LINUX_VERSION" == "Debian"  ] || \
 		 [ "$LINUX_VERSION" == "Kali"    ] || \
 		 [ "$LINUX_VERSION" == "Ubuntu"  ]; then
-		apt-get install -y libfreetype6-dev python3-dev python3-pip pkg-config
+		apt-get install -y libfreetype6-dev python3-dev pkg-config
+		if ! python3 -m pip --version; then
+			if apt-get install python3-pip; then
+				echo "INFO: Installed python3-pip via apt-get"
+			else
+				curl https://bootstrap.pypa.io/get-pip.py | python3
```
wolfthefallen: piped into `python3`. But cannot really confirm but all documentation from pypi this is the site and method to use if it is not available via a package manager. 

https://pip.pypa.io/en/stable/installing/
https://github.com/pypa/get-pip#usage

tools/install.sh
zeroSteiner: So it has to be on it's own line? Is it not available in Fedora 29?

wolfthefallen: correct

zeroSteiner: Okay can you note that then so we know that on `Fedora 29+` that may fail and it's fine. Is it fine though, can we still use the RPC terminal or is there a newer version of the package that we should be using?

wolfthefallen: I'll add a comment in the file. Not sure what package it is for Fedora 29+ but RPC Terminal is working just fine.

image

zeroSteiner:
Merged, did not test.

zeroSteiner:
So when testing with Fedora 29 (an upgraded from Fedora 28 installation) I get [this large stack trace](https://pastebin.com/WStKGXM2). It looks like it has something to do with PyYAML failing to build.

zeroSteiner:
Yes, I was able to fix it by running `pipenv install pyyaml==3.13`. Can you check which version of PyYAML you have, was 3.12 installing for you?

```
pipenv install pyyaml==3.13
Installing pyyaml==3.13…
Collecting pyyaml==3.13
Installing collected packages: pyyaml
Successfully installed pyyaml-3.13

Adding pyyaml to Pipfile's [packages]…
Pipfile.lock (ce08b2) out of date, updating to (af7b7a)…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…



Updated Pipfile.lock (ce08b2)!
Installing dependencies from Pipfile.lock (ce08b2)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 66/66 — 00:01:27
```

wolfthefallen:
I am 95% sure that is the version I installed during testing as that is the pinned version. I'll double check first thing in the morning before I start the rest of the release process.

PR #342, opened 2018-12-21
Url autocomplete

wolfthefallen:
Using the new graphql call to fetch webpage `.metadata.yml` file information. This PR takes that information to build valid urls available to the King Phisher server, and adds it to list that the entry web server url can reference to present to the end user for selection.

To test:
set up king phisher and place in the .metadata.yml in to your web directories to reference the landing pages you wish.

log in with king phisher client.. select campaign, 
* type in `h` all options should populate.
* type in the first couple of characters of a hostname, its options should populate

** if no hostname is available, it checks addresses available to the king phisher server and if it is not `0.0.0.0` it will present that as an option, else it will not populate anything for that landing page, as no valid uri could be made.

### DIFF
--- a/data/client/king_phisher/queries/get_site_templates.graphql
+++ b/data/client/king_phisher/queries/get_site_templates.graphql
@@ -1,5 +1,5 @@
 # retrieve information of all site templates
-query getsiteTemplates {
+query getSiteTemplates {
   siteTemplates {
     edges {
       node {
--- a/data/client/king_phisher/queries/get_site_templates.graphql
+++ b/data/client/king_phisher/queries/get_site_templates.graphql
@@ -1,0 +1,18 @@
+# retrieve information of all site templates
+query getsiteTemplates {
+  siteTemplates {
+    edges {
+      node {
+        created
+        hostname
+        path
+        metadata {
+          authors
+          classifiers
+          description
+          pages
+        }
+      }
+    }
+  }
+}
--- a/king_phisher/client/gui_utilities.py
+++ b/king_phisher/client/gui_utilities.py
@@ -102,6 +102,29 @@ def which_glade():
 	"""
 	return find.data_file(os.environ.get('KING_PHISHER_GLADE_FILE', 'king-phisher-client.ui'))
 
+def _store_extend(store, things, clear=False):
+	if clear:
+		store.clear()
+	for thing in things:
+		store.append(thing)
+
+def glib_idle_add_store_extend(store, things, clear=False, wait=False):
+	"""
+	Extend a GTK store object (either :py:class:`Gtk.ListStore` or :py:class:`Gtk.TreeStore`)
+	object using :py:func:`GLib.idle_add`.
+	This function is suitable for use in non-main GUI threads for synchronizing data.
+
+	note** regardless of *wait*, None is returned
+
+	:param store: The GTK storage object to add *things* to.
+	:type store: :py:class:`Gtk.ListStore`, :py:class:`Gtk.TreeStore`
+	:param tuple things: The array of things to add to *store*.
+	:param bool clear: Whether or not to clear the storage object before adding *things* to it.
+	:param bool wait: Whether or not to wait for the operation to complete before returning.
+	"""
+	idle_add = glib_idle_add_wait if wait else glib_idle_add_once
+	idle_add(_store_extend, store, things, clear)
+
 def glib_idle_add_once(function, *args, **kwargs):
 	"""
 	Execute *function* in the main GTK loop using :py:func:`GLib.idle_add`
--- a/king_phisher/client/tabs/mail.py
+++ b/king_phisher/client/tabs/mail.py
@@ -36,6 +36,7 @@ import hashlib
 import os
 import re
 import sys
+import time
 import urllib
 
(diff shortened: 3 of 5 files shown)


zeroSteiner:
Wow "** if no hostname is available, it checks addresses available to the king phisher server and if it is not 0.0.0.0 it will present that as an option, else it will not populate anything for that landing page, as no valid uri could be made.". That's a nice touch! :+1:

king_phisher/client/tabs/mail.py
```diff
@@ -866,6 +869,131 @@ def __init__(self, *args, **kwargs):
...
+		url_information = self._get_url_templates()
+		if not url_information:
+			return
+		for edge in url_information['siteTemplates']['edges']:
+			for page in edge['node']['metadata']['pages']:
+				urls = self._build_urls(edge['node']['hostname'], page, edge['node']['path'])
+				if not urls:
+					continue
+				for url in urls:
+					gui_utilities.glib_idle_add_once(self.url_list_store.append, [url])
```
zeroSteiner: So `Gtk.ListStore` doesn't have an `extend` method. I think this warrants a utility function to handle this in a single `glib_idle_add_*` call so we're not registering multiple one-shot routines to be done when it's idle. I'm surprised we haven't run into this elsewhere in the application yet.

I'm thinking something that would even handle the clear for you like this:
```python
# gui_utilities.py
def _store_extend(store, things):
    for thing in things:
        store.append(thing)

def glib_idle_add_store_extend(store, things, wait=False):
    """
    Extend a GTK store object (either :py:class:`Gtk.ListStore` or :py:class:`Gtk.TreeStore`) object using :py:func:`GLib.idle_add`. This function is suitable for use in non-main GUI threads for synchronizing data.

    :param store: The GTK storage object to add *things* to.
    :type store: :py:class:`Gtk.ListStore`, :py:class:`Gtk.TreeStore`
    :param tuple things: The array of things to add to *store*.
    :param bool wait: Whether or not to wait for the operation to complete before returning.
    """
    idle_add = glib_idle_add_wait if wait else glib_idle_add_once
    idle_add(_store_extend, store, things)
    # not that regardless of wait, None is returned
```

king_phisher/client/tabs/mail.py
```diff
@@ -866,6 +869,131 @@ def __init__(self, *args, **kwargs):
...
+
+	def _set_entry_completion_list_tsafe(self):
+		gui_utilities.glib_idle_add_once(self.url_list_store.clear)
+		url_information = self._get_url_templates()
+		if not url_information:
+			return
+		for edge in url_information['siteTemplates']['edges']:
+			for page in edge['node']['metadata']['pages']:
+				urls = self._build_urls(edge['node']['hostname'], page, edge['node']['path'])
+				if not urls:
```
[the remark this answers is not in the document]
wolfthefallen: unfortunately a `NoneType` is not iterable so it blows up :bomb: :fire: :boom: 
image

king_phisher/client/tabs/mail.py
```diff
@@ -936,50 +938,25 @@ def _webserver_url_key_press_event(self, widget, event):
...
+				gui_utilities.glib_idle_add_store_extend(self.url_list_store, urls)
+		self.logger.debug('successfully updated webserver url autocomplete liststore')
 		self.url_completion_last_load_time = time.time()
 
 	def _build_urls(self, hostname, page, path):
 		urls = []
 		if not hostname:
 			for address in self.config['server_config']['server.addresses']:
-				if address['host'] != '0.0.0.0':
+				if address['host'] != '0.0.0.0' and address['host'] != '::' and not ipaddress.ip_address(address['host'].is_link_local):
```
zeroSteiner: Sorry, but I was thinking about this some more and I think my original suggestion wasn't a good one. So for the case of `::` that's not the only representation of the same IPv6 address. I'm thinking normalizing to an IP address either way and doing `ip.is_global or ip.is_private and not ip.is_unspecified`. That way we'd complete global and private addresses, but not unspecified ones like `0.0.0.0` or other odd things I can't think of.

king_phisher/client/gui_utilities.py
```diff
@@ -102,6 +102,29 @@ def which_glade():
 	"""
 	return find.data_file(os.environ.get('KING_PHISHER_GLADE_FILE', 'king-phisher-client.ui'))
 
+def _store_extend(store, things, clear=False):
+	if clear:
+		store.clear()
+	for thing in things:
+		store.append(thing)
+
+def glib_idle_add_store_extend(store, things, clear=False, wait=False):
```
zeroSteiner: Can you add this function to `docs/source/king_phisher/client/gui_utilities.rst` so the docs include it?

data/client/king_phisher/queries/get_site_templates.graphql
```diff
@@ -0,0 +1,18 @@
+# retrieve information of all site templates
+query getsiteTemplates {
```
zeroSteiner: `query getSiteTemplates {` (capital S in Sites for camelCaseing).

zeroSteiner:
This is awesome, I'll have it landed here shortly. Excellent work @wolfthefallen.

PR #354, opened 2019-03-01
Updated CampaignAssistant

wolfthefallen:
Updated Campaign Assistant to provide advertised URLs from the king-phisher server and the ability to import KPM.

This PR adds one tab to the Campaign Assistant that displays the information from advertised landing pages from the King Phisher Server for the user to select when the are creating or configuring the campaign. In addition on the finalize page the user now has the option to import a campaign to quickly configure their message settings.

Now when you finalize on the Campaign Assistant, it will now emit `campaign-change` forcing a load of all the message settings that get set during the new functions in cooperated into the work flow. 

Test Cases:
- [ ] You can see the new tab `Advertised URLs`
  - [ ] If you select an entry it will load in the message configuration window when you close the Assistant
- [ ] Last page has a `Import` button to allow the user to import a KPM file
  - [ ] You select the KPM file and destination folder but does not load until assistant is 'Applied`
  - [ ] If you select an Advertised URL it over writes the URL supplied with the KPM file.
- [ ] GUI on the new tab is in line and flows with the rest of the King Phish GUI
- [ ] Labels and tool tips are clear and understandable

### DIFF
--- a/data/client/king_phisher/king-phisher-client.ui
+++ b/data/client/king_phisher/king-phisher-client.ui
@@ -1281,6 +1281,8 @@ Spencer McIntyre</property>
                             <property name="tooltip_text" translatable="yes">Destination folder for KPM contents.</property>
                             <property name="hexpand">True</property>
                             <property name="editable">False</property>
+                            <property name="primary_icon_stock">gtk-directory</property>
+                            <signal name="backspace" handler="signal_kpm_dest_folder" swapped="no"/>
                           </object>
                           <packing>
                             <property name="left_attach">0</property>
@@ -1308,8 +1310,10 @@ Spencer McIntyre</property>
                             <property name="hexpand">True</property>
                             <property name="editable">False</property>
                             <property name="invisible_char">●</property>
+                            <property name="primary_icon_stock">gtk-file</property>
                             <property name="primary_icon_activatable">False</property>
                             <property name="secondary_icon_activatable">False</property>
+                            <signal name="backspace" handler="signal_kpm_file_clear" swapped="no"/>
                           </object>
                           <packing>
                             <property name="left_attach">0</property>
@@ -1335,7 +1339,7 @@ Spencer McIntyre</property>
                             <property name="visible">True</property>
                             <property name="can_focus">True</property>
                             <property name="receives_default">True</property>
-                            <signal name="clicked" handler="signal_kpm_dest_folder" swapped="no"/>
+                            <signal name="clicked" handler="signal_kpm_dest_folder_clicked" swapped="no"/>
                           </object>
                           <packing>
                             <property name="left_attach">1</property>
--- a/data/client/king_phisher/king-phisher-client.ui
+++ b/data/client/king_phisher/king-phisher-client.ui
@@ -1203,13 +1203,181 @@ Spencer McIntyre</property>
         </child>
       </object>
       <packing>
-        <property name="title" translatable="yes">Advertised URLS</property>
+        <property name="title" translatable="yes">Target URL</property>
         <property name="complete">True</property>
         <property name="has_padding">False</property>
       </packing>
     </child>
     <child>
       <object class="GtkBox" id="CampaignAssistant.page7">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="margin_top">5</property>
+        <property name="orientation">vertical</property>
+        <property name="spacing">10</property>
+        <child>
+          <object class="GtkLabel" id="CampaignAssistant.label_kpm_top_level">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="halign">center</property>
+            <property name="label" translatable="yes">KPM Import</property>
+            <attributes>
+              <attribute name="weight" value="bold"/>
+              <attribute name="scale" value="1.5"/>
+            </attributes>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="position">0</property>
+          </packing>
+        </child>
+        <child>
+          <object class="GtkGrid" id="grid8">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="margin_left">3</property>
+            <property name="margin_right">3</property>
+            <property name="row_spacing">3</property>
+            <property name="column_spacing">10</property>
+            <property name="row_homogeneous">True</property>
+            <child>
+              <object class="GtkFrame" id="CampaignAssistant.frame_kpm_import">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="label_xalign">0</property>
+                <property name="shadow_type">none</property>
+                <child>
+                  <object class="GtkAlignment">
+                    <property name="visible">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="top_padding">5</property>
+                    <property name="bottom_padding">5</property>
+                    <property name="left_padding">12</property>
+                    <property name="right_padding">5</property>
+                    <child>
+                      <object class="GtkGrid" id="CampaignAssistant.grid_credential_validation_regexs1">
+                        <property name="visible">True</property>
+                        <property name="can_focus">False</property>
+                        <property name="row_spacing">3</property>
+                        <property name="column_spacing">3</property>
+                        <child>
+                          <object class="GtkLabel" id="CampaignAssistant.label_kpm_file">
+                            <property name="visible">True</property>
+                            <property name="can_focus">False</property>
+                            <property name="halign">start</property>
+                            <property name="label" translatable="yes">Message Configuration File</property>
+                          </object>
+                          <packing>
+                            <property name="left_attach">0</property>
+                            <property name="top_attach">0</property>
+                            <property name="width">2</property>
+                          </packing>
+                        </child>
+                        <child>
+                          <object class="GtkEntry" id="CampaignAssistant.entry_kpm_dest_folder">
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                            <property name="tooltip_text" translatable="yes">Destination folder for KPM contents.</property>
+                            <property name="hexpand">True</property>
+                            <property name="editable">False</property>
+                          </object>
+                          <packing>
+                            <property name="left_attach">0</property>
+                            <property name="top_attach">3</property>
+                          </packing>
+                        </child>
+                        <child>
+                          <object class="GtkLabel" id="CampaignAssistant.label_kpm_dest_folder">
+                            <property name="visible">True</property>
+                            <property name="can_focus">False</property>
+                            <property name="halign">start</property>
+                            <property name="label" translatable="yes">Destination Folder</property>
+                          </object>
+                          <packing>
+                            <property name="left_attach">0</property>
+                            <property name="top_attach">2</property>
+                            <property name="width">2</property>
+                          </packing>
+                        </child>
+                        <child>
+                          <object class="GtkEntry" id="CampaignAssistant.entry_kpm_file">
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                            <property name="tooltip_text" translatable="yes">File path to KPM file</property>
+                            <property name="hexpand">True</property>
+                            <property name="editable">False</property>
+                            <property name="invisible_char">●</property>
+                            <property name="primary_icon_activatable">False</property>
+                            <property name="secondary_icon_activatable">False</property>
+                          </object>
+                          <packing>
+                            <property name="left_attach">0</property>
+                            <property name="top_attach">1</property>
+                          </packing>
+                        </child>
+                        <child>
+                          <object class="GtkButton" id="CampaignAssistant.button_select_kpm_file">
+                            <property name="label" translatable="yes">Select</property>
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                            <property name="receives_default">True</property>
+                            <signal name="clicked" handler="signal_kpm_select_clicked" swapped="no"/>
+                          </object>
+                          <packing>
+                            <property name="left_attach">1</property>
+                            <property name="top_attach">1</property>
+                          </packing>
+                        </child>
+                        <child>
+                          <object class="GtkButton" id="CampaignAssistant.button_select_kpm_dest_folder">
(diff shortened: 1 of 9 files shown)


king_phisher/client/tabs/mail.py
```diff
@@ -1243,6 +1244,7 @@ def signal_kpc_campaign_set(self, *_):
 		context_id = self.status_bar.get_context_id('campaign name')
 		self.status_bar.pop(context_id)
 		self.status_bar.push(context_id, self.config['campaign_name'])
+		#self.go
```
zeroSteiner: This can be removed.

king_phisher/client/gui_utilities.py
```diff
@@ -513,6 +514,35 @@ def gtk_widget_destroy_children(widget):
 	for child in widget.get_children():
 		child.destroy()
 
+def set_listbox_classifiers(label, listbox, classifiers):
```
zeroSteiner: You wrote the documentation but forgot to add it to the `docs/source/king_phisher/client/gui_utilities.rst` file for it to be included in the build.

data/client/king_phisher/queries/get_site_template.graphql
```diff
@@ -0,0 +1,18 @@
+# retrieve information of all site templates
```
zeroSteiner: `# retrieve information for all site templates`

king_phisher/client/assistants/campaign.py
```diff
@@ -30,18 +30,33 @@
...
+from king_phisher.client.widget import extras
 from king_phisher.client.widget import resources
 from king_phisher.client.widget import managers
 
+from gi.repository import Gtk
 import advancedhttpserver
 
 __all__ = ('CampaignAssistant',)
 
+_ModelNamedRow = collections.namedtuple('ModelNamedRow', (
```
zeroSteiner: Excellent use of the newer, better way!

king_phisher/client/assistants/campaign.py
```diff
@@ -127,6 +153,37 @@ def __init__(self, application, campaign_id=None):
...
+		self.domain_completion.set_model(self._domain_list_store)
+		self.domain_completion.set_text_column(0)
+		self.gobjects['entry_domain_filter'].set_completion(self.domain_completion)
+
+		self.kpm_target_file = None
+		self.kpm_dest_dir = None
+
+		tvm = managers.TreeViewManager(self.gobjects['treeview_url_selector'])
+		tvm.set_column_titles(
+			['Domain', 'Landng Page', 'Tags', 'URL'],
```
zeroSteiner: Is the third entry tags or classifiers? Should try to keep the verbiage consistent.

zeroSteiner:
Received this stack trace while testing this out:
```
Error Type: builtins.KeyError
Error Details: KeyError('name')
Error UID: 67a28e1d-6e9e-47ff-a0e8-e4dd03cdbad3
RPC Error: N/A (Not an RPC error)
King Phisher Version: 1.13.0-beta3 (rev: 6597cd4eeacf)
Platform Version: Linux: Fedora 29 Twenty Nine
Python Version: 3.7.2
Gtk Version: 3.24.1
Timezone: America/New_York

Thread Information:
===================
    pydevd.Writer (alive=True daemon=True)
    pydevd.Reader (alive=True daemon=True)
    pydevd.CommandThread (alive=True daemon=True)
    Thread-7 (alive=True daemon=True)

Stack Trace:
============
  File "/home/steiner/Repositories/king-phisher/king_phisher/client/assistants/campaign.py", line 526, in signal_assistant_apply
    self.config['campaign_name'] = properties['name']
KeyError: 'name'
```

data/client/king_phisher/king-phisher-client.ui
```diff
@@ -946,7 +946,279 @@ Spencer McIntyre</property>
...
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="position">4</property>
+          </packing>
+        </child>
+      </object>
+      <packing>
+        <property name="title" translatable="yes">Advertised URLS</property>
```
zeroSteiner: Can you change this to `Target URL`?

king_phisher/client/assistants/campaign.py
zeroSteiner: Yeah, this should be classifiers. Can you also please swap it with URL so it's last? Since the classifiers are so long, the URL is pushed off the screen and since that's what I want to set, it would be helpful if it were featured more predominately.

king_phisher/client/assistants/campaign.py
```diff
@@ -144,6 +201,35 @@ def is_editing_campaign(self):
...
+		return self._url_thread is None or not self._url_thread.is_alive()
+
+	@property
+	def _server_uses_ssl(self):
+		return any(address['ssl'] for address in self.config['server_config']['server.addresses'])
+
+	def signal_expander_activate(self, expander):
+		return
+
+	def signal_kpm_import_clicked(self, _):
```
zeroSteiner: Instead of a giant button for this, can you put it on it's own tab in the assistant? Just a simple tab right after the URL selection with an entry and a spot to open the file selection dialog like what exists in the Mailer Configuration tab.

zeroSteiner:
Also getting this stack trace. To reproduce it, delete your `campaign_id` and `campaign_name` keys from your configuration tab. This will prompt you to select or create a new campaign the next time you login. Now when you do this, go to create a new Campaign and you should get this stack trace.

```
Exception in thread Thread-7:
  File "/usr/lib64/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/home/steiner/Repositories/king-phisher/king_phisher/utilities.py", line 526, in run
    super(Thread, self).run()
  File "/usr/lib64/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/steiner/Repositories/king-phisher/king_phisher/client/tabs/mail.py", line 939, in _set_entry_completion_list_tsafe
    url_information = self.application.rpc.graphql_find_file('get_site_templates.graphql')
AttributeError: 'NoneType' object has no attribute 'graphql_find_file'
```

king_phisher/client/assistants/campaign.py
```diff
@@ -30,18 +30,49 @@
...
+	'hostname',
+	'page',
+	'url',
+	'classifiers',
+	'authors',
+	'description',
+	'created'
+))
+
+class KpmImport(object):
```
zeroSteiner: This whole class is to store redundant information. You should just retrieve these two values from the the entries when you need them. Move the `KpmImport.check` method into `CampaignAssistant.set_kpm_assistant_complete`. I originally thought this was just going to be an immutable namedtuple for convenience but what you're doing instead is initializing it once in `CampaignAssistant.__init__` and then manually syncing the entry data over into it which is no bueno.

king_phisher/client/assistants/campaign.py
```diff
@@ -144,6 +220,52 @@ def is_editing_campaign(self):
...
+	@property
+	def _url_thread_is_ready(self):
+		return self._url_thread is None or not self._url_thread.is_alive()
+
+	@property
+	def _server_uses_ssl(self):
+		return any(address['ssl'] for address in self.config['server_config']['server.addresses'])
+
+	def set_kpm_assistant_complete(self):
+		if not self._import_kpm.check():
```
zeroSteiner: All four of these lines could just be `self.assistant.set_page_complete(self.assistant.get_nth_page(self.assistant.get_current_page()), self._import_kpm.check())` `KpmImport.check` needs to be moved though. If you wanted the lines to be shorter, it would be reasonable to do:
```
page = self.assistant.get_nth_page(self.assistant.get_current_page())
self.assistant.set_page_complete(page, self._import_kpm.check())
```.

Also, in general it's bad practice to invert the logic with `not` and then include an else block. Just drop the inversion and swap the blocks.

File: king_phisher/client/assistants/campaign.py [hunk 10]
```diff
@@ -144,6 +220,52 @@ def is_editing_campaign(self):
...
+		dialog = extras.FileChooserDialog('Destination Directory', self.parent)
+		response = dialog.run_quick_select_directory()
+		dialog.destroy()
+		if not response:
+			return False
+		self._import_kpm.dest_folder = response['target_path']
+		self.gobjects['entry_kpm_dest_folder'].set_text(self._import_kpm.dest_folder)
+		self.set_kpm_assistant_complete()
+
+	def signal_kpm_file_clear(self, _):
```
zeroSteiner: Once you get rid of the `KpmImport` class and pull this information at runtime, you should combine this and the `signal_kpm_dest_folder` handlers into a single method like:

```
def signal_kpm_entry_clear(self, entry_widget):
    entry_widget.set_text('')
    self.set_kpm_assistant_complete()
```

king_phisher/client/assistants/campaign.py
```diff
@@ -156,6 +278,61 @@ def _set_comboboxes(self):
...
+
+		domains = []
+		self._url_classifiers = {}
+		for edge in url_information['siteTemplates']['edges']:
+			for page in edge['node']['metadata']['pages']:
+				if hostname:
+					if not edge['node']['hostname'].startswith(hostname):
+						continue
+				domains.append(edge['node']['hostname'])
+				self._url_classifiers[edge['node']['hostname']] = edge['node']['metadata']['classifiers']
```
zeroSteiner: This logic seems buggy. The hostname is not enough to uniquely identify this set of classifiers because multiple templates could share the same hostname. You should use the `(edge['node']['hostname'], page)` tuple or the url as the key to ensure you're not overwriting classifiers from other templates.

You're also calculating the `_url_classifiers` when you load the raw data, so that one at least makes alot of sense to me.

king_phisher/client/assistants/campaign.py
```diff
@@ -156,6 +278,61 @@ def _set_comboboxes(self):
...
+		# make domain list unique in case multiple pages are advertised for the domains
+		domains = [[domain] for domain in set(domains)]
+		gui_utilities.glib_idle_add_store_extend(self._domain_list_store, domains)
+
+	def _build_url(self, hostname, page):
+		if not hostname:
+			for address in self.config['server_config']['server.addresses']:
+				ip = ipaddress.ip_address(address['host'])
+				if not ip.is_unspecified and (ip.is_global or ip.is_private):
+					hostname = address['host']
```
zeroSteiner: This should break out of the loop once you've found a match.

king_phisher/client/assistants/campaign.py
```diff
@@ -156,6 +278,61 @@ def _set_comboboxes(self):
...
+		if not hostname:
```
zeroSteiner: Once you break out of the loop, this would be a great use case to use an [else clause](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops).

```
if not hostname:
    for address in self.config['server_config']['server.addresses']:
        ip = ipaddress.ip_address(address['host'])
        if not ip.is_unspecified and (ip.is_global or ip.is_private):
            hostname = address['host']
            break
    else:
        hostname = 'localhost'

File: king_phisher/client/assistants/campaign.py [hunk 14]
```diff
@@ -364,6 +559,22 @@ def signal_assistant_apply(self, _):
...
 
 		self.application.emit('campaign-changed', cid)
+
+		old_cid = self.config['campaign_id']
+		self.config['campaign_id'] = cid
+		self.config['campaign_name'] = properties['name']
+
+		if self._import_kpm.check():
+			if not self.application.main_tabs['mailer'].emit('message-data-import', self._import_kpm.target_file, self._import_kpm.dest_folder):
+				gui_utilities.show_dialog_info('Failure', self.parent, 'Failed to imported the message configuration.')
```
zeroSteiner: `'Failed to import the message...` not `'Failed to imported the...`

king_phisher/client/assistants/campaign.py
```diff
@@ -127,6 +172,37 @@ def __init__(self, application, campaign_id=None):
 			self.gobjects['label_intro_body'].set_text('This assistant will walk you through creating and configuring a new King Phisher campaign.')
 			self.gobjects['label_intro_title'].set_text('New Campaign')
 
+		self._url_selected = None
+		self._url_classifiers = {}
+		self._url_thread = None
+
+		self.domain_completion = Gtk.EntryCompletion()
```
zeroSteiner: It doesn't look like you need to keep a reference for this on `self`.

king_phisher/client/assistants/campaign.py
```diff
@@ -127,6 +172,37 @@ def __init__(self, application, campaign_id=None):
...
+				Gtk.CellRendererText(),
+				Gtk.CellRendererText(),
+				Gtk.CellRendererText(),
+				Gtk.CellRendererText(),
+			]
+		)
+		self._url_model = Gtk.TreeStore(str, str, str, str, str, str, str)
+		self._url_model.set_sort_column_id(0, Gtk.SortType.ASCENDING)
+		self.gobjects['treeview_url_selector'].set_model(self._url_model)
+		self._url_information = {
```
zeroSteiner: You're adding 5 `_url_*` attributes. Could you combine some into this `_url_information` dictionary, `selected` and `classifiers` both seem like they'd fit. You'd then have the raw data, the created timestamp, the selection and the classifiers (which came from the raw data) all together.

king_phisher/client/assistants/campaign.py
```diff
@@ -262,6 +439,23 @@ def _get_graphql_company(self, company_name):
 		}""", {'name': company_name})
 		return results['db']['company']
 
+	def signal_url_treeview_row_activated(self, treeview, path, column):
+		model_row = self._url_model[path]
+		self._set_info(model_row)
+
+	def _set_info(self, model_row):
```
zeroSteiner: This should be a bit more descriptive since we have other `_set_info` methods. Rename it to `_set_info_url_details`

king_phisher/client/assistants/campaign.py
```diff
@@ -127,6 +172,37 @@ def __init__(self, application, campaign_id=None):
 			self.gobjects['label_intro_body'].set_text('This assistant will walk you through creating and configuring a new King Phisher campaign.')
 			self.gobjects['label_intro_title'].set_text('New Campaign')
 
+		self._url_selected = None
```
zeroSteiner: This shouldn't need to be stored. You should get the selected row from the treeview when you need the value.

PR #372, opened 2019-06-06
Access method

Nib0x62s:
# Access Method

## Description
There seems to be a significant need for determining the permissions a user has
on files outside of the current user. Access acts as a wrapper for **os.access** by
extending the ability to specify a user and/or group along with a file to test permissions against.

## Requirements
- [ ] Allow users to specify a file to test against
- [ ] Allow users to specify a mode to test against (R_OK, W_OK, X_OK)
- [ ] Enable user to specify either a user *or* a group to test against.

## Run / test configuration
* **Run Config**
  * access(*test-file*, *mode*, user=*user*, group=*group*)
  * Either user or group must be specified only one can be left as default.
* **Owner Tests**
  * Make a user the owner: `chown <user> test-file`
    * Owner and read `chmod u+r test-file`
    * Owner and write `chmod u+w test-file`
    * Owner and execute `chmod u+x test-file`
* **User's Group Tests & All group Test**
  * Make a user's group the owner: `chown :<user's-group> test-file`
  * To test other groups other than the user's group, enumerate a group they're apart of <br> (e.g, adm) and add that group to the file: `chown :<enumerated-group> test-file`
    * Group and read `chmod g+r test-file`
    * Group and write `chmod g+w test-file`
    * Group and execute `chmod g+x test-file`
* **Other Tests**
  * Other and read `chmod o+r test-file`
  * Other and write `chmod o+w test-file`
  * Other and execute `chmod o+x test-file`

### DIFF
--- a/king_phisher/server/fs_utilities.py
+++ b/king_phisher/server/fs_utilities.py
@@ -34,9 +34,12 @@ import grp
 import os
 import pwd
 
+import collections
+import king_phisher.server.pylibc as pylibc
 import smoke_zephyr.utilities
 
 from king_phisher import constants
+from os import stat
 
 def chown(path, user=None, group=constants.AUTOMATIC, recursive=True):
 	"""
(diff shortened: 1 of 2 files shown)


king_phisher/server/fs_utilities.py
```diff
@@ -83,3 +86,77 @@ def chown(path, user=None, group=constants.AUTOMATIC, recursive=True):
...
+		:py:class:`~king_phisher.constants.AUTOMATIC`, the group that *user*
+		belongs too will be used.
+	:type group: int, str, ``None``, :py:class:`~king_phisher.constants.AUTOMATIC`
+	:return: Returns ``True`` only if the user or group has the mode of permission specified else returns ``False``.
+	:rtype: bool
+	"""
+	if (user is constants.AUTOMATIC or user is None) and (group is constants.AUTOMATIC or group is None):
+		raise ValueError('either user or group must be specified')
+
+	if isinstance(user, str):
```
zeroSteiner: So instead of handling user and group seperately like you do, you should "resolve" the user and group arguments to numeric uid and gid values like the `chown` function does. I recommend you reuse that code by placing it in a private function (prefix it with `_`).

That way you can consolidate the logic and move on to (in english):
1. Is the owner the UID (resolved from *user*) I'm looking for and does the owner have the permissions (the *mode* argument) I'm looking for?
1. Is the group a GID I'm looking for and does the group have the permissions I'm looking for?
    * GID checks are a bit more tricky, if `group` is `constants.AUTOMATIC`, then you should check **all the groups that the user is a member of**, if `group` is `None` skip checking the group altogether, if `group` is an explicit number then only check that one
1. Check if everyone has the access I'm looking for.

PR #374, opened 2019-06-07
SSH Forwarding Update

Nib0x62s:
# SSH Forwarding Update

## Description
There has been a recurring problem with users running into issues setting their
preferred ssh-key in the client configuration. Spencer brought up the idea of having
king-phisher attempt to authenticate with all possible keys so users would no longer
have to specify their preferred ssh-key mitigating this recurring problem. This update will attempt to authenticate will all available ssh-keys. 

## Test-Cases
- [x] Successfully authenticate to king-phisher without specifying a preferred-ssh-key. 
- [ ] Correctly log all attempted key's while authenticating.

### DIFF
--- a/king_phisher/ssh_forward.py
+++ b/king_phisher/ssh_forward.py
@@ -153,8 +153,10 @@ class SSHTCPForwarder(threading.Thread):
 			else:
 				self.logger.warning('failed to identify the user specified key for ssh authentication')
 
-		if not self.__connected and len(agent_keys) == 1:
-			self.__try_connect(look_for_keys=False, pkey=agent_keys[0])
+		if not self.__connected and agent_keys:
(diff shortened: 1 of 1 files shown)


zeroSteiner:
Did we confirm that this is what OpenSSH does by default?

Nib0x62s:
Yeah, I documented the research and posted it in the slack channel.

zeroSteiner:
Merged! Only made one minor change in commit e471e20c9e164e239a7dde9d64d2c1511226da6e to add a change log line, and log the type of authentication that is being attempted. Decided against logging each keys to avoid potentially including sensitive info in log output, so only the fact that the agent-provided keys are being attempted and how many there are is included.

Thanks @Nib0x62s !

PR #375, opened 2019-06-11
Access Method

Nib0x62s:
# Access Method

## Description
There seems to be a significant need for determining the permissions a user has
on files outside of the current user. Access acts as a wrapper for **os.access** by
extending the ability to specify a user and/or group along with a file to test permissions against.

## Requirements
- [x] Allow users to specify a file to test against
- [x] Allow users to specify a mode to test against (R_OK, W_OK, X_OK)
- [x] Enable user to specify either a user *or* a group to test against.
- [x] Successfully enumerate a users or groups permission on a specific filet.

## Run / test configuration
* **Run Config**
  * access(*test-file*, *mode*, user=*user*, group=*group*)
  * Either user or group must be specified only one can be left as default.
* **Owner Tests**
  * Make a user the owner: `chown <user> test-file`
    * Owner and read `chmod u+r test-file`
    * Owner and write `chmod u+w test-file`
    * Owner and execute `chmod u+x test-file`
* **User's Group Tests & All group Test**
  * Make a user's group the owner: `chown :<user's-group> test-file`
  * To test other groups other than the user's group, enumerate a group they're apart of <br> (e.g, adm) and add that group to the file: `chown :<enumerated-group> test-file`
    * Group and read `chmod g+r test-file`
    * Group and write `chmod g+w test-file`
    * Group and execute `chmod g+x test-file`
* **Other Tests**
  * Other and read `chmod o+r test-file`
  * Other and write `chmod o+w test-file`
  * Other and execute `chmod o+x test-file`

### DIFF
--- a/king_phisher/server/fs_utilities.py
+++ b/king_phisher/server/fs_utilities.py
@@ -134,6 +134,25 @@ def access(path, mode, user=constants.AUTOMATIC, group=constants.AUTOMATIC):
 	uid, gid = _resolve_target_ids(user, group)
 	file_info = os.stat(path)
 
+	# If there are no permissions to check for then yes the user has no more than no permissions.
+	if not mode:
+		return True
+
+	# Other checks
+	# start with no test flags
+	test_mode = 0
+	# then check if R_OK is to be checked for
+	if mode & os.R_OK:
+		# and if so, add the proper flag to be tested
+		test_mode |= stat.S_IROTH
+	if mode & os.W_OK:
+		test_mode |= stat.S_IWOTH
+	if mode & os.X_OK:
+		test_mode |= stat.S_IXOTH
+	if file_info.st_mode & test_mode:
+		return True
+
+	# User Checks
 	if file_info.st_uid == uid:
 		# start with no test flags
 		test_mode = 0
@@ -148,7 +167,24 @@ def access(path, mode, user=constants.AUTOMATIC, group=constants.AUTOMATIC):
 		if file_info.st_mode & test_mode:
 			return True
 
-	if isinstance(user, str) and group == constants.AUTOMATIC:
+	# Group checks
+	if file_info.st_gid == gid:
+		test_mode = 0
+		# then check if R_OK is to be checked for
+		if mode & os.R_OK:
+			# and if so, add the proper flag to be tested
+			test_mode |= stat.S_IRUSR
+		if mode & os.W_OK:
+			test_mode |= stat.S_IWUSR
+		if mode & os.X_OK:
+			test_mode |= stat.S_IXUSR
+		if file_info.st_mode & test_mode:
+			return True
+
+	# If there were no groups specified then enumerate all of the users gids and test them all.
+	if group == constants.AUTOMATIC:
+		if not isinstance(user, str):
+			user = pylibc.getpwuid(user).pw_name
 		# check all the groups
 		if any(file_info.st_gid == gid for gid in pylibc.getgrouplist(user)):
 			test_mode = 0
--- a/king_phisher/server/fs_utilities.py
+++ b/king_phisher/server/fs_utilities.py
@@ -31,6 +31,8 @@
 #
 
 import os
+import stat
+
 import smoke_zephyr.utilities
 
 from king_phisher import constants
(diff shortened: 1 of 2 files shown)


king_phisher/server/fs_utilities.py
```diff
@@ -56,6 +58,72 @@ def chown(path, user=None, group=constants.AUTOMATIC, recursive=True):
...
+	"""
+	This is a high-level wrapper around :py:func:`os.access` to provide
+	additional functionality. Similar to `os.access` this high-level wrapper
+	will test the given path for a variety of access modes. Additionally however,
+	*user* or *group* can be specified to test against a specific of user or group.
+
+	.. versionadded:: 1.14.0
+
+	:param str path: The path to test access for.
+	:param str mode: The mode to test access for. Set to `R_OK` to test for readability,
```
zeroSteiner: So mode should be an integer and should support multiple flag values like `os.R_OK | os.W_OK` instead of being a string.

king_phisher/server/fs_utilities.py
```diff
@@ -56,6 +58,72 @@ def chown(path, user=None, group=constants.AUTOMATIC, recursive=True):
...
+		group_permissions = stat.S_IXGRP
+		other_permissions = stat.S_IXOTH
+
+	if file_info.st_uid == user_id and file_info.st_mode & user_permissions:
+		return True
+
+	if file_info.st_gid == group and file_info.st_mode & group_permissions:
+		return True
+
+	for group_id in pylibc.getgrouplist(user):
```
zeroSteiner: This should only check all groups if `group == constants.AUTOMATIC`, other than that this logic looks relatively sound.

king_phisher/server/fs_utilities.py
```diff
@@ -56,6 +58,72 @@ def chown(path, user=None, group=constants.AUTOMATIC, recursive=True):
...
+		belongs too will be used.
+	:type group: int, str, ``None``, :py:class:`~king_phisher.constants.AUTOMATIC`
+	:return: Returns ``True`` only if the user or group has the mode of permission specified else returns ``False``.
+	:rtype: bool
+	"""
+
+	user_id, group = _resolve_information(user, group)
+	file_info = os.stat(path)
+
+	if mode == 'R_OK':
```
zeroSteiner: If you're going to define, user, group, and other permissions this way, then the values should be initialized to 0 and then you should set the bits based on the mode flags like:
```python
user_permissions = 0
group_permissions = 0
other_permissions = 0
if mode & os.R_OK:
    user_permissions |= stat.S_IRUSR
    group_premissions |= stat.S_IRGRP
    other_permissions |= stat.S_IROTH
if mode & os.W_OK:  # note that this is `if` not `elsif` so flags can be `os.R_OK | os.W_OK`
    user_permissions |= stat.S_IWUSR
    # and so on and so forth
```

Then later when you go to check them, you'd need to make sure all the bits are equal doing:
```python
(file_info.st_mode & user_permissions == user_permissions)
```
This will ensure that all the flags set in `user_permissions` are present and not just one of them (which would be enough to cause the result to be non-zero and return true. This will still have the short coming of requiring that the permissions are all present for either the user, group or other but I think that's fine so long as we document it.

king_phisher/server/fs_utilities.py
```diff
@@ -56,6 +58,72 @@ def chown(path, user=None, group=constants.AUTOMATIC, recursive=True):
...
+	if file_info.st_gid == group and file_info.st_mode & group_permissions:
+		return True
+
+	for group_id in pylibc.getgrouplist(user):
+		if group_id == file_info.st_gid and file_info.st_mode & group_permissions:
+			return True
+
+	if file_info.st_mode & other_permissions:
+		return True
+
```
zeroSteiner: Technically if `user_permissions`, `group_permissions` and `other_permissions` are all 0, then this function would return false when it should return true. This would be an edge case where the user doesn't specify anything in mode, implying a check without any permission requirements, in which case "yes the user has no or more than no permissions" so return True.

king_phisher/server/fs_utilities.py
```diff
@@ -107,3 +106,59 @@ def chown(path, user=None, group=constants.AUTOMATIC, recursive=True):
...
+			# and if so, add the proper flag to be tested
+			test_mode |= stat.S_IRUSR
+		if mode & os.W_OK:
+			test_mode |= stat.S_IWUSR
+		if mode & os.X_OK:
+			test_mode |= stat.S_IXUSR
+		if file_info.st_mode & test_mode:
+			return True
+
+	if isinstance(user, str) and group == constants.AUTOMATIC:
```
zeroSteiner: If `group == constants.AUTOMATIC` and `user` is not a string, but `uid` was resolved, then you should resolve the username for that `uid`. Definitely check if you already have the string to avoid another call to libc if you can, but otherwise you'll have to resolve it yourself.

If `group` is `constants.AUTOMATIC` and `uid` is not an integer, then something horrible happened.

zeroSteiner: Forgot to mention, if necessary resolve the `uid` to a username with [`pylibc.pwgetuid`](https://king-phisher.readthedocs.io/en/latest/king_phisher/server/pylibc.html#king_phisher.server.pylibc.getpwuid).

zeroSteiner:
Merged after successfully testing this. Thanks alot @Nib0x62s!

I made some minor changes to consolidate the flag generate and checking logic in commit b02fffa43f32c10f3e5f82c3423f45f816d7c0c8.

PR #377, opened 2019-06-19
Proxy requests

Nib0x62s:
# Proxy Requests 

This PR will allow users to specify a HTTP_PROXY to use through the king_phisher configuration. 
Utilizing the HTTPS_PROXY environment variable king-phisher will now support the ability to update the variable during run-time through the client configuration GUI.  

- [x] Allow users to update the HTTPS_PROXY environment variable through the client configuration GUI.
- [x] When the client loads king-phisher will pull in what ever the value of HTTPS_PROXY and updte th config appropriately. 
     - Test this by doing `HTTPS_PROXY=http://me:password@host.com:8080/ ./KingPhisher -L DEBUG`
- [x] When a user does Edit -> Preferences -> Client to look at the proxy settings KP should load whatever is in the config and properly display the information in the respected text fields.


zeroSteiner:
@Nib0x62s Since we talked about this out of band a little bit just hit me up when it's ready for review.

data/client/king_phisher/client_config.json
```diff
@@ -68,5 +69,5 @@
   "ssh_server": "localhost:22",
   "ssh_username": "",
   "text_font": "monospace 11",
-  "text_source.theme": "cobalt"
+  "text_source.theme": "cobalt",
```
zeroSteiner: I think this is technically invalid JSON. You can't have a trailing comma, but King Phisher makes a best effort to address this type of issue to be user friendly. I'd prefer to keep the files that ship with the source conforming to the spec though.

king_phisher/client/application.py
```diff
@@ -540,6 +540,8 @@ def do_config_load(self, load_defaults):
 		Load the client configuration from disk and set the
 		:py:attr:`~.KingPhisherClientApplication.config` attribute.
 
+		Check the proxy environment variable and set the appropriately.
```
zeroSteiner: "...and set **them** appropriately."

king_phisher/client/application.py
```diff
@@ -552,6 +554,13 @@ def do_config_load(self, load_defaults):
 			for key, value in client_template.items():
 				if not key in self.config:
 					self.config[key] = value
+			env_proxy = os.environ['HTTPS_PROXY']
```
zeroSteiner: This will raise a `KeyError` if `HTTPS_PROXY` is not set. Instead use `os.environ.get('HTTPS_PROXY')`.

king_phisher/client/application.py
```diff
@@ -552,6 +554,13 @@ def do_config_load(self, load_defaults):
...
+			if env_proxy is not None:
+				if env_proxy != self.config['proxy.url']:
+					self.logger.warning('Proxy environment variable does not match the user configuration.'
```
zeroSteiner: The indentation is off here. We actually have a note in the [style guide](https://king-phisher.readthedocs.io/en/latest/development/style_guide.html#multi-line-indentation) about how it should be done for consistency. Also just about every other log line is not a full and complete sentence but instead starts with a lower case letter (except for acronyms) and does not end with a period.

king_phisher/client/application.py
```diff
@@ -552,6 +554,13 @@ def do_config_load(self, load_defaults):
...
+										'Setting configuration appropriately')
+					self.config['proxy.url'] = env_proxy
+			os.environ['HTTPS_PROXY'] = self.config['proxy.url']
```
zeroSteiner: This should also set the `os.environ['HTTP_PROXY']` value too.

king_phisher/client/dialogs/configuration.py
```diff
@@ -274,6 +279,31 @@ def interact(self):
...
 
+	def save_proxy_settings(self):
+		proxy_url = urllib.parse.urlparse(self.gtk_builder_get('entry_proxy_username').get_text())
+		proxy_username = self.gtk_builder_get('entry_proxy_username').get_text()
+		proxy_password = self.gtk_builder_get('entry_proxy_password').get_text()
+		if not proxy_url.hostname and not proxy_url.scheme:
+			gui_utilities.show_dialog_warning('Invalid Proxy Values', self.parent,
+												'The proxy settings you have entered are not valid.')
+			return
+		if proxy_username and proxy_password and proxy_url.port:
```
zeroSteiner: You can convert the `ParseResult` returned by `urllib.parse.urlparse` with `urllib.parse.urlunparse` instead of using a format string.

king_phisher/client/dialogs/configuration.py
```diff
@@ -274,6 +279,31 @@ def interact(self):
...
+		elif proxy_url.port:
+			formatted_proxy_url = utilities.nonempty_string('{0}//{1}:{2}/'.format(proxy_url.scheme, proxy_url.hostname, proxy_url.port))
+		elif proxy_username and proxy_password:
+			formatted_proxy_url = utilities.nonempty_string('{0}//{3}/'.format(proxy_url.scheme, proxy_username,
+																							proxy_password, proxy_url.hostname, proxy_url.port))
+		else:
+			formatted_proxy_url = utilities.nonempty_string('{0}//{1}/'.format(proxy_url.scheme, proxy_url.hostname))
+
+		self.config['proxy.url'] = formatted_proxy_url
+		if proxy_url.group('protocol').lower() == 'http':
```
zeroSteiner: Here we want to set both HTTP_PROXY and HTTPS_PROXY to the same value whether it's an http or https URL.

king_phisher/client/dialogs/configuration.py
```diff
@@ -274,6 +279,31 @@ def interact(self):
 		self.dialog.destroy()
 		return response
 
+	def save_proxy_settings(self):
+		proxy_url = urllib.parse.urlparse(self.gtk_builder_get('entry_proxy_username').get_text())
+		proxy_username = self.gtk_builder_get('entry_proxy_username').get_text()
+		proxy_password = self.gtk_builder_get('entry_proxy_password').get_text()
+		if not proxy_url.hostname and not proxy_url.scheme:
+			gui_utilities.show_dialog_warning('Invalid Proxy Values', self.parent,
```
zeroSteiner: This logic looks like it'll throw a warning if no proxy is specified. Since this is an optional setting we should only throw a warning if the value is incomplete. For example they specify `http://` (schema but no hostname) or they specify `test.com` (hostname but no schema).

The minimum things we would need would be the schema and hostname. Port can be inferred. You should also check that the schema is either `http` or `https`.

king_phisher/client/dialogs/configuration.py
```diff
@@ -265,6 +269,7 @@ def interact(self):
 		response = self.dialog.run()
```
zeroSteiner: It doesn't look like you have the logic to set the applicable GTK entries. Before `dialog.run()` you should parse the URL, and then populate the URL, username and password GTK entry widgets.

king_phisher/client/application.py
```diff
@@ -552,6 +555,17 @@ def do_config_load(self, load_defaults):
...
 					self.config[key] = value
+			env_proxy = os.environ.get('HTTPS_PROXY')
+			if env_proxy is not None:
+				proxy_url = urllib.parse.urlparse(env_proxy)
+				if proxy_url.hostname and not proxy_url.scheme or proxy_url.scheme and not proxy_url.hostname:
+					self.logger.error('invalid proxy url scheme or hostname')
+					return
+				if env_proxy != self.config['proxy.url']:
+					self.logger.warning('proxy environment variable does not match the user configuration')
+					self.config['proxy.url'] = env_proxy
```
zeroSteiner: We would want this set either way. Whether `env_proxy` matches it or not. If it does match it, logging an `info` message would be good while still logging a `warning` message if it does not match like you are doing.

king_phisher/client/application.py
```diff
@@ -552,6 +555,17 @@ def do_config_load(self, load_defaults):
 			for key, value in client_template.items():
 				if not key in self.config:
 					self.config[key] = value
+			env_proxy = os.environ.get('HTTPS_PROXY')
+			if env_proxy is not None:
+				proxy_url = urllib.parse.urlparse(env_proxy)
+				if proxy_url.hostname and not proxy_url.scheme or proxy_url.scheme and not proxy_url.hostname:
```
zeroSteiner: You could simplify this to no `if not (proxy_url.hostname and proxy_url.scheme)` and I think it'd be a bit more readable too.

king_phisher/client/application.py
```diff
@@ -552,6 +555,17 @@ def do_config_load(self, load_defaults):
...
+					self.logger.error('invalid proxy url scheme or hostname')
```
zeroSteiner: `self.logger.error('invalid proxy url (missing scheme or hostname)')`

king_phisher/client/dialogs/configuration.py
```diff
@@ -32,7 +32,10 @@
 
 import collections
 import logging
+import os
 import string
+import urllib.parse
+
```
zeroSteiner: Don't need a second blank line here per the [Style Guide](https://king-phisher.readthedocs.io/en/latest/development/style_guide.html#style-guide).

king_phisher/client/dialogs/configuration.py
```diff
@@ -274,6 +290,22 @@ def interact(self):
 		self.dialog.destroy()
 		return response
 
+	def save_proxy_settings(self):
+		proxy_url = urllib.parse.urlparse(self.gtk_builder_get('entry_proxy_url').get_text())
+		proxy_username = self.gtk_builder_get('entry_proxy_username').get_text()
+		proxy_password = self.gtk_builder_get('entry_proxy_password').get_text()
+		if proxy_url.hostname and not proxy_url.scheme or proxy_url.scheme and not proxy_url.hostname:
```
zeroSteiner: Simplify again with `if proxy_url and not (proxy_url.hostname and proxy_url.scheme)`. This would evaluate to true when anything is set in the proxy_url but the hostname or scheme is missing.

king_phisher/client/dialogs/configuration.py
```diff
@@ -274,6 +290,22 @@ def interact(self):
 		self.dialog.destroy()
 		return response
 
+	def save_proxy_settings(self):
+		proxy_url = urllib.parse.urlparse(self.gtk_builder_get('entry_proxy_url').get_text())
```
zeroSteiner: Would be a good idea to strip the whitespace from the URL and username at least. Passwords I guess could technically have whitespace so I'll leave that up to you but it would be a bit more forgiving on input from the user in cases where maybe they're pasting the value in from some document.

king_phisher/client/dialogs/configuration.py
```diff
@@ -274,6 +291,27 @@ def interact(self):
...
+	def save_proxy_settings(self):
+		proxy_url = urllib.parse.urlparse(self.gtk_builder_get('entry_proxy_url').get_text().strip())
+		proxy_username = self.gtk_builder_get('entry_proxy_username').get_text().strip()
+		proxy_password = self.gtk_builder_get('entry_proxy_password').get_text().strip()
+		if proxy_url and not (proxy_url.hostname and proxy_url.scheme):
+			gui_utilities.show_dialog_warning('Invalid Proxy Settings', self.parent, 'The proxy url you have submitted is not valid.')
+			return
+		try:
+			proxy_url.port
+		except:
```
zeroSteiner: This should except the specific exception class that should be raised, in this case that would be a `ValueError`.

king_phisher/client/dialogs/configuration.py
```diff
@@ -274,6 +291,27 @@ def interact(self):
...
+			gui_utilities.show_dialog_warning('Invalid Port', self.parent, 'The port must be an integer.')
+			return
+		netloc = proxy_url.netloc
+		if proxy_username and proxy_password:
```
zeroSteiner: Just check the `proxy_username` here to allow users to use proxies with users that don't have a password.

king_phisher/client/dialogs/configuration.py
```diff
@@ -274,6 +291,27 @@ def interact(self):
...
+		proxy_url = urllib.parse.urlparse(self.gtk_builder_get('entry_proxy_url').get_text().strip())
+		proxy_username = self.gtk_builder_get('entry_proxy_username').get_text().strip()
+		proxy_password = self.gtk_builder_get('entry_proxy_password').get_text().strip()
+		if proxy_url and not (proxy_url.hostname and proxy_url.scheme):
+			gui_utilities.show_dialog_warning('Invalid Proxy Settings', self.parent, 'The proxy url you have submitted is not valid.')
+			return
+		try:
+			proxy_url.port
+		except:
+			gui_utilities.show_dialog_warning('Invalid Port', self.parent, 'The port must be an integer.')
```
zeroSteiner: This is a good warning dialog, but you should use the same title as the previous one "Invalid Proxy Settings". That way their logically grouped together.

king_phisher/client/application.py
```diff
@@ -552,6 +555,19 @@ def do_config_load(self, load_defaults):
...
+			proxy_url = urllib.parse.urlparse(env_proxy)
+			if not (proxy_url.hostname and proxy_url.scheme):
+				self.logger.error('invalid proxy url (missing scheme or hostname)')
+				return
+			if env_proxy != self.config['proxy.url']:
+				self.logger.warning('proxy environment variable does not match the user configuration')
+			else:
+				self.logger.info('proxy environment variable is the same as the user configuration')
+			self.config['proxy.url'] = env_proxy
+		os.environ['HTTPS_PROXY'] = self.config['proxy.url']
```
zeroSteiner: This is crashing on me when I don't have an environment variable already set. So if `self.config['proxy.url']` is `None`, this will crash because you can't set environment variables to `None`.

You might need another conditional statement to set the `HTTP(S)_PROXY` environment variables when `self.config['proxy.url']` is set. That should come after the check you're already done. If neither are set, then I would highly recommend deleting the `HTTP_PROXY` environment variable so we know that the configuration is synced and that particular value won't get leaked in.

This is what I mocked up until I got to the second error.

```
		if env_proxy is not None:
			#... do what you're doing already but here
		elif self.config['proxy.url']:
			os.environ['HTTPS_PROXY'] = self.config['proxy.url']
			os.environ['HTTP_PROXY'] = self.config['proxy.url']
		else:
                        # delete HTTP_PROXY if there is not a known configuration
			os.environ.pop('HTTP_PROXY', None)
```

king_phisher/client/dialogs/configuration.py
```diff
@@ -221,6 +223,19 @@ def _configure_settings_plugins(self):
 		for plugin_klass in plugin_klasses:
 			self._configure_settings_plugin_options(plugin_klass)
 
+	def _configure_settings_proxy(self):
+		formatted_proxy_url = urllib.parse.urlparse(self.config['proxy.url'])
```
zeroSteiner: This is also crashing for me when `self.config['proxy.url']` is None. To reproduce it just start King Phisher without a proxy configuration then open `Edit > Preferences`.

This is the stack trace:
```
Error Type: builtins.TypeError
Error Details: TypeError('Must be string, not bytes')
Error UID: 592845cb-a8a3-4324-84bb-8e84aaa41458
RPC Error: N/A (Not an RPC error)
King Phisher Version: 1.14.0-beta1 (rev: dad7205fa100)
Platform Version: Linux: Fedora 30 Thirty
Python Version: 3.7.3
Gtk Version: 3.24.8
Timezone: America/New_York

Thread Information:
===================
    pydevd.Writer (alive=True daemon=True)
    pydevd.Reader (alive=True daemon=True)
    pydevd.CommandThread (alive=True daemon=True)
    Thread-23 (alive=True daemon=True)
    Thread-20 (alive=True daemon=False)
    RPCAsyncWorker (alive=True daemon=False)
    Thread-27 (alive=True daemon=True)
    Thread-28 (alive=True daemon=True)
    Dummy-56 (alive=True daemon=True)
    Dummy-57 (alive=True daemon=True)

Stack Trace:
============
  File "/home/steiner/Repositories/king-phisher/king_phisher/client/windows/main.py", line 120, in signal_activate_edit_preferences
    self.application.show_preferences()
  File "/home/steiner/Repositories/king-phisher/king_phisher/client/application.py", line 779, in show_preferences
    if dialog.interact() != Gtk.ResponseType.CANCEL:
  File "/home/steiner/Repositories/king-phisher/king_phisher/client/dialogs/configuration.py", line 277, in interact
    self._configure_settings_proxy()
  File "/home/steiner/Repositories/king-phisher/king_phisher/client/dialogs/configuration.py", line 237, in _configure_settings_proxy
    self.gtk_builder_get('entry_proxy_url').set_text(proxy_url)
TypeError: Must be string, not bytes
```

The issue is that when `self.config['proxy.url']` is `None`, the members of the `ParseResult` are bytes, which is some straight up crazy talk but hey I didn't write this API. :man_shrugging: 

```
In [2]: urllib.parse.urlparse('test')                                                                                                                                                                               
Out[2]: ParseResult(scheme='', netloc='', path='test', params='', query='', fragment='')

In [3]: urllib.parse.urlparse(None)                                                                                                                                                                                 
Out[3]: ParseResultBytes(scheme=b'', netloc=b'', path=b'', params=b'', query=b'', fragment=b'')  
```

Anyways if `self.config['proxy.url']` is None, you can safely skip setting the entries since they'll default to blank anyways.

PR #379, opened 2019-06-26
Downgraded matplotlib

Nib0x62s:
Matplotlib 2.2.4 is the latest version of the library that supports python 3.4 thus a downgrade is needed.

- Ran `pipenv uninstall matplotlib`
- Ran `pipenv install matplotlib==2.2.4` to install the correct version.
- Manually edited requirements.txt to respect the new version.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,7 +20,6 @@ graphql-relay = "==0.4.5"
 icalendar = "==4.0.3"
 ipaddress = "==1.0.22"
 jsonschema = "==2.6.0"
-matplotlib = "==3.0.2"
 msgpack-python = "==0.5.6"
 paramiko = "==2.4.2"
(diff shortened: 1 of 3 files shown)


zeroSteiner:
Testing steps

- [ ] Test that graphs are generating correctly
- [ ] This will work with the Windows build
- [ ] The legends show up in the generated graphs<sup>1</sup>

<sup>1</sup> This [patch](https://github.com/securestate/king-phisher/commit/e42bcd75b6325c83d4f6f3c565920e90919386ab) was a temp fix that removed the legends when MPL was v3.0.2. It does *not* look like the changes need to be undone.

wolfthefallen:
All tested and good; cleaned up conflict and landed

PR #380, opened 2019-06-27
Removed pip-file

Nib0x62s:
- [x] Pipfile.lock is generated by install.sh when deleted from a fresh clone from github.com/securesate/king-phisher
- [x] Pipfile.lock is generated on an older version of kp after running the update step's as detailed in the wiki.
- [x] Pipfile.lock is generated on an older version of kp after running the update step's as detailed in the wiki _even when pipfile.lock is deleted from master_

## To verify:
1. Run `git clone https://github.com/securestate/king-phisher/`
2. Remove pipfile.lock while on the **master** branch.
3. Checkout any older version e.g. `git checkout v1.3.0` 
4. run `git fetch origin && git pull origin master`
5. Now running `tools/install.sh --update` should pull down the correct pipfile.lock


wolfthefallen:
I can confirm that by deleting the Pipefile.lock will correctly generate and update accordingly based off the Pipefile when installing and updating. This will correct the issue when packages have a variation of hash between python interpreter versions.

PR #381, opened 2019-06-27
Pipenv `--three`

wolfthefallen:
This PR forces python3 to be used during the installation of the `pipenv`.
In addition updated change log to reflect the delectation of the `Pipefile.lock` file and added the file to `.gitignore` file.

Testing:
- [x] remove python3 `pipenv` installation and install `pipenv` with python2
- [x] `./KingPhisher --env-install` installs default python3 interpreter for the system.
- [x] `./KingPhisher --env-update` updates pip requirements and default python3 interpreter for the system.

### DIFF
--- a/docs/source/change_log.rst
+++ b/docs/source/change_log.rst
@@ -15,8 +15,8 @@ Version 1.14.0
 
 * Added the ``Message-ID`` MIME header to outgoing messages
 * Attempt SSH authentication with all agent-provided SSH keys
-* Deleted `Pipefile.lock` from repository to prevent hash issues between python interpreter versions.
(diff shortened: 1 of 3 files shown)


docs/source/change_log.rst
```diff
@@ -15,6 +15,8 @@ Version 1.14.0
 
 * Added the ``Message-ID`` MIME header to outgoing messages
 * Attempt SSH authentication with all agent-provided SSH keys
+* Deleted `Pipefile.lock` from repository to prevent hash issues between python interpreter versions.
```
zeroSteiner: So this isn't markdown, to mark something as a literal it should be two back ticks like Message-ID before. Also Python should be capitalized and the local convention here is to not use periods.

Same comments for the line below.

PR #386, opened 2019-07-08
Server auto install plugin pip requirements

wolfthefallen:
# Server auto install pip requirements for plugins
This PR addresses `pipenv` removing python packages manually being installed that are not in the `Pipefile`. 

This PR adds two new option into the server config. First is the `server.plugin_library_path` where the user can define where they want plugin libraries to install. This value is defaulted to: `/var/king-phisher/libs/plugin_site_packages`. The second value is `server.plugin_install_requirements` where the user can specify if King Phisher Server should automatically try to install missing package dependencies of it's plugins. This value defaults to `True`.

## Testing
Set your testing environment to use a plugin that requires a dependency that is not included in the King Phisher Environment.
- [x] Set`server.plugin_install_requirements` option to false. Start server and it does not install package and fails out for not having the package installed.
- [x] Set`server.plugin_install_requirements` to True. Start Server and verify it auto installs python package into path defined by server config.
- [x] Profit?

### DIFF
--- a/data/server/king_phisher/server_config.yml
+++ b/data/server/king_phisher/server_config.yml
@@ -107,6 +107,10 @@ server:
   #  some_plugin:
   #    # An option specific to this particular plugin
   #    value_for_some_plugin: true
+  # Plugin Library depenacy path
+  plugin_library_path: /var/king-phisher/libs/plugin_site_packages
+  # Auto install server plugin requirements
+  plugin_install_requirements: True
 
   # Require an ID associated with a campaign to load pages
   require_id: true
--- a/data/server/king_phisher/server_config.yml
+++ b/data/server/king_phisher/server_config.yml
@@ -101,16 +101,16 @@ server:
   # Add additional search paths for plugin directories
   #plugin_directories:
   #  - /some/plugin/path
+  # Auto install server plugin python requirements
+  plugin_install_requirements: True
+  # Plugin library dependency path
+  plugin_library_path: /var/king-phisher/libs/plugin_site_packages
   # List the plugins that need to be enabled
   #plugins:
   #  # The plugin to enable's name
   #  some_plugin:
   #    # An option specific to this particular plugin
   #    value_for_some_plugin: true
-  # Plugin Library depenacy path
-  plugin_library_path: /var/king-phisher/libs/plugin_site_packages
-  # Auto install server plugin requirements
-  plugin_install_requirements: True
 
   # Require an ID associated with a campaign to load pages
   require_id: true
--- a/king_phisher/client/windows/plugin_manager.py
+++ b/king_phisher/client/windows/plugin_manager.py
@@ -391,7 +391,15 @@ class PluginManagerWindow(gui_utilities.GladeGObject):
 				self._update_status_bar_tsafe(
 					"Installing {:,} dependenc{} for plugin {} from PyPi.".format(len(packages), 'y' if len(packages) == 1 else 'ies', named_row.title)
 				)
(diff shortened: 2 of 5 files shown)


data/server/king_phisher/server_config.yml
```diff
@@ -107,6 +107,10 @@ server:
   #  some_plugin:
   #    # An option specific to this particular plugin
   #    value_for_some_plugin: true
+  # Plugin Library depenacy path
```
zeroSteiner: These settings should go above `#plugins:` so they're alphabetized. Should go `plugin_directories`, `plugin_install_requirements`, and then `plugin_library_path`.

king_phisher/plugins.py
```diff
@@ -607,6 +610,19 @@ def load_module(self, name, reload_module=False):
 			recursive_reload(module)
 		return module
 
+	def install_packages(self, packages):
```
zeroSteiner: This method should be documented and be sure to remember to add the `.. versionadded::` specifier.

king_phisher/server/plugins.py
```diff
@@ -134,24 +134,25 @@ def __init__(self, config):
...
 			klass = self[plugin]
 			for req_type, req_value, req_met in klass.compatibility:
 				req_type = req_type.lower()
 				if req_met:
 					self.logger.debug("plugin '{0}' requirement {1} ({2}) met".format(plugin, req_type, req_value))
 					continue
-				self.logger.warning("plugin '{0}' unmet requirement {1} ({2})".format(plugin, req_type, req_value))
-				raise errors.KingPhisherPluginError(plugin, 'failed to meet requirement: ' + req_type)
-			# enable the plugin
+				if req_type == 'Required Package' or not req_met:
```
zeroSteiner: With the line `req_type = req_type.lower()` above, the `req_type == 'Required Package'` portion of this expression will never evaluate to True.

king_phisher/server/plugins.py
```diff
@@ -134,24 +134,25 @@ def __init__(self, config):
...
+					self.logger.warning("plugin '{0}' unmet requirement {1} ({2})".format(plugin, req_type, req_value))
+					if config.get_if_exists('server.plugin_library_path') and config.get_if_exists('server.plugin_install_requirements'):
+						self.install_packages([req_value])
```
zeroSteiner: You should check the exit status from pip here, log a warning / error message as appropriate and skip the reloading if the installation failed.

king_phisher/client/windows/plugin_manager.py
```diff
@@ -400,7 +391,7 @@ def _plugin_install_tsafe(self, catalog_model, repo_model, model_row, named_row)
 				self._update_status_bar_tsafe(
 					"Installing {:,} dependenc{} for plugin {} from PyPi.".format(len(packages), 'y' if len(packages) == 1 else 'ies', named_row.title)
 				)
-				pip_results = self._pip_install(packages)
+				pip_results = self.application.plugin_manager.install_packages(packages)
```
zeroSteiner: You should either catch the exception that `install_packages` might raise or check that `plugin_manager.library_path` is set before calling this.

zeroSteiner:
All tests checked out. I made a few minor grammatical changes, and set the default path in the config to `/var/king-phisher/lib`. After being "resolved" the final path is `/var/king-phisher/lib/pythonX.Y/site-packages` which is pretty ideal IMHO.

I'll have this landed in a moment, thanks @wolfthefallen!

PR #387, opened 2019-07-11
Allow users to delete proxy setings via gui

Nib0x62s:
This PR will allow user's to unset their proxy settings via the GUI instead of throwing invalid proxy settings. 

# Testing Steps:
- [x] Load up king phisher and set the proxy through preferences. 
- [x] Clear out the text field and click apply.
- [x] Reloading king-phisher or the preferences tab should verify that the proxy settings have be unset.


zeroSteiner:
Looks good, thanks @Nib0x62s. Only change is one I'll make and that's to use `None` instead of an empty string to show that the proxy is not defined. I'll have this landed shortly.

zeroSteiner:
This has been merged but because I had to rebase it, GitHub won't show it.

PR #388, opened 2019-07-11
Updated to a more explicit error message for TOTP

Nib0x62s:
# New explicit error message for TOTP.

### Testing:
- [ ] Load up the KP client
- [ ] Sign in using valid credentials
- [ ] Enter an invalid TOTP (e.g., nuFsiI&R)


zeroSteiner:
This should be fixed in the TOTP self enrollment plugin [over here](https://github.com/securestate/king-phisher-plugins/tree/master/client/totp_enrollment). If it fails to log the user in, and it is the time sync issue then that would be a server side issue and the user that is trying to log in may not even have the permissions necessary to fix it.

Nib0x62s:
grr

PR #398, opened 2019-08-23
Update Pipfile and pinned versions

wolfthefallen:
This PR updates the Pipefile requirements and pinned versions documentation.

Removed the below packages as they are requested by other packages and dont need to be pinned:
removed - python-dateutil as it is required by Alembic, icalendar
removed - numpy as it is required by matplotlib and basemaps


zeroSteiner:
Landed thanks @wolfthefallen

PR #400, opened 2019-09-06
Windows updates for plugin pypi packages installation

wolfthefallen:
This PR adds in the python.exe and pip.exe to the Windows build to allow the frozen windows application to install python packages from pypi during plugin installation. Also added in `disutils`, `six`, `os`, `sys`, to allow pip to build packages. 

Added in logic to prevent installation of packages that are not supported on the operating system.

### DIFF
--- a/king_phisher/client/windows/plugin_manager.py
+++ b/king_phisher/client/windows/plugin_manager.py
@@ -866,7 +866,7 @@ class PluginManagerWindow(gui_utilities.GladeGObject):
 			return
 		if plugin_id in self.application.plugin_manager:
 			klass = self.application.plugin_manager[plugin_id]
-			compatibility_details = klass.compatibility
+			compatibility_details = list(klass.compatibility)
 		else:
 			repo_model, catalog_model = self._get_plugin_model_parents(self._last_plugin_selected)
 			compatibility_details = list(self.catalog_plugins.compatibility(catalog_model.id, repo_model.id, named_plugin.id))
@@ -884,7 +884,6 @@ class PluginManagerWindow(gui_utilities.GladeGObject):
 		row = 0
 		for row, req in enumerate(compatibility_details):
 			grid.insert_row(row)
-			self.logger.debug(req)
 			if row == 0:
 				label = Gtk.Label(req[0])
 			else:
--- a/king_phisher/client/windows/plugin_manager.py
+++ b/king_phisher/client/windows/plugin_manager.py
@@ -311,10 +311,20 @@ class PluginManagerWindow(gui_utilities.GladeGObject):
 					version=plugin_info['version'],
 					visible_enabled=True,
 					visible_installed=True,
-					sensitive_installed=True,
+					sensitive_installed=self._is_sensitive_installed(self.catalog_plugins.compatibility(catalog.id, repo.id, plugin_name)),
 					type=_ROW_TYPE_PLUGIN
 				))
 		gui_utilities.glib_idle_add_once(self.__store_add_node, catalog_node)
+
+	def _is_sensitive_installed(self, compatibility_details):
+		for detail in compatibility_details:
+			if detail.type.name == 'platforms' and not detail.comptabile:
+				return False
+			elif detail.type.name == 'minimum-python-version' and not detail.comptabile:
+				return False
+			elif detail.type.name == 'minimum-kp-version' and not detail.comptabile:
+				return False
+		return True
 
 	def _get_plugin_model_parents(self, plugin_model_row):
 		return _ModelNamedRow(*plugin_model_row.parent), _ModelNamedRow(*plugin_model_row.parent.parent)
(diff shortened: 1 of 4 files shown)


king_phisher/client/windows/plugin_manager.py
```diff
@@ -357,22 +394,25 @@ def _plugin_install(self, model_row):
 					return
 		self._worker_thread_start(self._plugin_install_tsafe, catalog_model, repo_model, model_row, named_row)
 
+	def _dialog_error_tsafe(self, error_message):
```
zeroSteiner: Can you please put this back? The reason why this was originally inlined instead of a method is that the title was always  `Failed To Install` but the method name makes it sound more flexible than that. That is to say that this method shouldn't be called elsewhere to show other errors that don't have to do with the installation process. This should either take the title as an argument or go back to being inlined using partial like it was.

king_phisher/plugins.py
```diff
@@ -135,6 +136,15 @@ class OptionString(OptionBase):
 	"""A plugin option which is represented with a string value."""
 	pass
 
+Requirement = collections.namedtuple('Requirement', ('type', 'value', 'comptabile'))
+
+_RequirementMeta = collections.namedtuple('RequirementMeta', ('name', 'title'))
+class _RequirementType(enum.Enum):
+	PACKAGE = _RequirementMeta(name='package', title='Required Package')
+	PYTHON_VERSION = _RequirementMeta('minimum-python-version', 'Minimum Python Version')
+	VERSION = _RequirementMeta('minimum-kp-version', 'Minimum King Phisher Version')
```
zeroSteiner: Please change the `type` here to `minimum-version`. I know it's a little ambiguous and in hind sight it should have been king-phisher-version but to be consistent with the source key, let's keep it at `minimum-version`.

king_phisher/plugins.py
```diff
@@ -189,20 +199,20 @@ def compatibility_iter(self):
 		if self._storage.get('minimum-python-version'):
 			# platform.python_version() cannot be used with StrictVersion because it returns letters in the version number.
 			available = StrictVersion(self._storage['minimum-python-version']) <= StrictVersion('.'.join(map(str, sys.version_info[:3])))
-			yield ('Minimum Python Version', self._storage['minimum-python-version'], available)
+			yield Requirement(_RequirementType.PYTHON_VERSION, self._storage['minimum-python-version'], available)
```
zeroSteiner: Change the string here from `...self._storage['minimum-python-version']...` to `...self._storage[_RequirementType.PYTHON_VERSION.name]` to keep that string usage limited.

king_phisher/plugins.py
```diff
@@ -135,6 +136,15 @@ class OptionString(OptionBase):
 	"""A plugin option which is represented with a string value."""
 	pass
 
+Requirement = collections.namedtuple('Requirement', ('type', 'value', 'comptabile'))
+
+_RequirementMeta = collections.namedtuple('RequirementMeta', ('name', 'title'))
+class _RequirementType(enum.Enum):
```
zeroSteiner: This should be public because you need it in `_compatibility_to_dict` in king_phisher/client/windows/plugin_manager.py.

king_phisher/client/windows/plugin_manager.py
```diff
@@ -339,13 +350,39 @@ def _plugin_enable(self, model_row):
...
 		self.config['plugins.enabled'].append(named_row.id)
 
+	def _compatibility_to_dict(self, compatibility_details):
+		required_packages = []
+		kp_version = None
+		python_version = None
+		supported_os = None
+
+		for detail in compatibility_details:
+			if detail[0].lower() == 'required package':
```
zeroSteiner: Use the `RequirementType` enum you defined here. Should be able to simply do `detail == RequirementType.PYTHON_VERSION` etc.

king_phisher/client/windows/plugin_manager.py
```diff
@@ -339,13 +350,39 @@ def _plugin_enable(self, model_row):
 		self._set_model_item(model_row.path, enabled=True, sensitive_installed=False)
 		self.config['plugins.enabled'].append(named_row.id)
 
+	def _compatibility_to_dict(self, compatibility_details):
```
zeroSteiner: I don't even see where this function is referenced, if it's not used please remove the whole thing.

king_phisher/client/windows/plugin_manager.py
```diff
@@ -310,11 +311,21 @@ def _add_catalog_to_tree_tsafe(self, catalog):
...
 					visible_installed=True,
-					sensitive_installed=True,
+					sensitive_installed=self._is_sensitive_installed(self.catalog_plugins.compatibility(catalog.id, repo.id, plugin_name)),
 					type=_ROW_TYPE_PLUGIN
 				))
 		gui_utilities.glib_idle_add_once(self.__store_add_node, catalog_node)
 
+	def _is_sensitive_installed(self, compatibility_details):
+		for detail in compatibility_details:
+			if detail.type.name == 'platforms' and not detail.comptabile:
```
zeroSteiner: Compare `detail.type` to the actual `RequirementType` enum after making it public.

wolfthefallen:
Closing this PR until I Iron out a windows build issue with it.

Issue #415, opened 2019-11-21
King Phisher 1.15.0 fails to build pipenv 


True-Demon:
## Issue Description
Using the king-phisher installer script, the install fails to build the pipenv properly, causing dependency issues and possibly further issues with the server. This appears to only be an issue with king-phisher v1.15.0+ (including 1.16.0-beta)

## Reproduction Steps
Run `wget -q https://github.com/securestate/king-phisher/raw/master/tools/install.sh && sudo bash ./install.sh` on a clean kali-light-amd64 system.

## Environment Details

Host OS: Linux kingphisher1 5.2.0-kali2-amd64 #1 SMP Debian 5.2.9-2kali1 (2019-08-22) x86_64 GNU/Linux
King Phisher Version: Version: 1.15.0 (rev: 0be34a20b0f3)

## Error Details / Stack Trace
```
root@kingphisher1:/opt# wget -q https://github.com/securestate/king-phisher/raw/master/tools/install.sh && \
> sudo bash ./install.sh
INFO: Linux version detected as Kali
Install and use PostgreSQL? (Highly recommended and required for upgrading) [Y/n] y
INFO: Will install and configure PostgreSQL for the server
INFO: Attempting to update apt-get cache package information
Get:1 http://kali.download/kali kali-rolling InRelease [30.5 kB]
Get:2 http://kali.download/kali kali-rolling/main amd64 Packages [16.5 MB]
Get:3 http://kali.download/kali kali-rolling/non-free amd64 Packages [191 kB]
Get:4 http://kali.download/kali kali-rolling/contrib amd64 Packages [104 kB]
Fetched 16.9 MB in 3s (6,565 kB/s)                     
Reading package lists... Done
INFO: Downloading and installing the King Phisher server to /opt/king-phisher
INFO: Successfully cloned the git repo
INFO: Synchronizing Kali OS dependencies
Reading package lists... Done
Building dependency tree       
Reading state information... Done
libfreetype6-dev is already the newest version (2.10.1-2).
pkg-config is already the newest version (0.29-6).
python3-dev is already the newest version (3.7.5-1).
0 upgraded, 0 newly installed, 0 to remove and 129 not upgraded.
pip 19.3.1 from /usr/local/lib/python3.7/dist-packages/pip (python 3.7)
Reading package lists... Done
Building dependency tree       
Reading state information... Done
gir1.2-gtk-3.0 is already the newest version (3.24.12-1).
[...]

PR #416, opened 2019-11-21
Added Numpy Pipfile requirement + version ref reason

True-Demon:
Discovered a rare bug when numpy is missing from the native environment due to basemap failing to check whether numpy is installed prior to importing it in its setup.py file. 

The current workaround is to require numpy before matplotlib and basemap are installed so as to guarantee it will succeed in all cases.

This resolves issue #415

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -44,6 +44,7 @@ MarkupSafe = "==1.1.1"
 PyYAML = "==5.1.2"
 SQLAlchemy = "==1.3.8"
 XlsxWriter = "==1.2.1"
+numpy = "==1.16.4" 
 basemap = {file = "https://github.com/matplotlib/basemap/archive/v1.2.1rel.tar.gz"}
 pyproj = "==2.4.0"
 rule-engine = "~=1.1"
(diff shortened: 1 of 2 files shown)


zeroSteiner:
Looks good, I'll give this a test later today and see about landing it. Thanks alot @True-Demon !

zeroSteiner:
Quick question actually, have you verified that this fixes the issue on a fresh install? My concern is that if we're running into an issues with the random ordering of dependency installation that we'd still have the same problem if basemaps is installed *before* numpy is explicitly installed. When you manually install numpy you're installing it before everything else and by extension basemaps, but I'm not convinced that when it's done automatically from the Pipfile, that numpy will still be installed first, thus fixing the issue.

True-Demon:
In my research, I discovered that Pip does not provide a means to specify an order of installation for pipfile package dependencies, and there is currently nothing to suggest this feature is going to be implemented in the near future. The only thing I can say for certain is that the installation will succeed after no more than two attempts when numpy is missing because numpy will be installed in the first round. The only time an installation will take more than one try is if pip tries to install matplotlib/basemap before it installs numpy.

Unfortunately, I think this is the best workaround we can provide. :(

zeroSteiner:
That issue is an excellent find, it really clarifies the behavior we're seeing here. Good job!

Yeah so with this being the current state of things, I'd agree that we should implement this as a change for the time being. I'll look at submitting a PR upstream to basemaps to have numpy explicitly included as a dependency.

PR #450, opened 2020-11-20
Fix incorrect URL

g0tmi1k:
Wrong subdomain for kali-docs.

...Though it may be worth removing, as Kali 2 hasn't been supported since January 2016.


zeroSteiner:
Merged, thanks for the update!

g0tmi1k:
🎉

PR #476, opened 2021-11-12
adding empty alt text to tracking_dot_image_tag

elreydetoda:
Just a small commit to add an empty alt tag, because websites like https://www.mail-tester.com/ mark that as a thing to improve on if you don't have alt tag at least present (even if it is empty).

Alternatively, another solution would be to allow the user to pass a parameter to specify an alt text like how `inline_image` does, but that was a lot more work :upside_down_face: . Feel free to close it if you'd like to go an alternative route, I made the modification to my client and figured I suggest it in case anyone else thought it was useful.


zeroSteiner:
Seems legit, thanks for the contribution.

elreydetoda:
anytime, thanks for the awesome project!

